PR #1, opened 2016-07-19
Set up standalone Docker and Travis

lukasjuhrich:
TODO
- [x] Remove comments
- [x] Move `&&` position
- [x] Don't introduce separate layer for installation of bower
- [x] Use tmpfs for postgres container
- [x] Improve bower initialization

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -11,3 +11,6 @@ doc/_build
 .vagrant
 web/static/libs/**
 
+*.log
+.coverage
+TAGS
--- a/.travis.yml
+++ b/.travis.yml
@@ -30,5 +30,4 @@ before_install:
 
 script:
   - docker-compose build
-  - docker-compose run --rm -u root web pip install coverage pylint
-  - docker-compose run --rm web nosetests --with-xunit --with-coverage --cover-erase --cover-branches --cover-package=web,pycroft,legacy
+  - docker-compose run --rm -u root web sh -c 'pip install coverage pylint && su pycroft -c "nosetests --with-xunit --with-coverage --cover-erase --cover-branches --cover-package=web,pycroft,legacy"'
--- a/.travis.yml
+++ b/.travis.yml
@@ -29,5 +29,6 @@ before_install:
   - "docker-compose version"
 
 script:
-  - docker-compose build
-  - docker-compose run --rm -u root web sh -c 'pip install coverage pylint && su pycroft -c "nosetests --with-xunit --with-coverage --cover-erase --cover-branches --cover-package=web,pycroft,legacy"'
+  - export DOCKER_COMPOSE="docker-compose -f docker-compose.bare.yml"
+  - $DOCKER_COMPOSE build
+  - $DOCKER_COMPOSE run --rm -u root web sh -c 'pip install coverage pylint && su pycroft -c "nosetests --with-xunit --with-coverage --cover-erase --cover-branches --cover-package=web,pycroft,legacy"'
--- a/.travis.yml
+++ b/.travis.yml
@@ -29,6 +29,6 @@ before_install:
   - "docker-compose version"
 
 script:
-  - export DOCKER_COMPOSE="docker-compose -f docker-compose.bare.yml"
+  - export DOCKER_COMPOSE="docker-compose -f docker-compose-travis.yml"
   - $DOCKER_COMPOSE build
   - $DOCKER_COMPOSE run --rm -u root web sh -c 'pip install coverage pylint && su pycroft -c "nosetests --with-xunit --with-coverage --cover-erase --cover-branches --cover-package=web,pycroft,legacy"'
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,27 +8,14 @@ cache:
 
 services:
   - docker
-env:
-  global:
-    - DOCKER_VERSION=1.11.1-0~trusty
-    - DOCKER_COMPOSE_VERSION=1.7.1
 
 before_install:
   # list docker-engine versions
   - apt-cache madison docker-engine
-
-  # upgrade docker-engine to specific version
-  - sudo apt-get -o Dpkg::Options::="--force-confnew" install -y docker-engine=${DOCKER_VERSION}
-
-  # upgrade docker-compose
-  - sudo rm /usr/local/bin/docker-compose
-  - curl -L https://github.com/docker/compose/releases/download/${DOCKER_COMPOSE_VERSION}/docker-compose-`uname -s`-`uname -m` > docker-compose
-  - chmod +x docker-compose
-  - sudo mv docker-compose /usr/local/bin
   - "docker version"
   - "docker-compose version"
 
 script:
   - export DOCKER_COMPOSE="docker-compose -f docker-compose-travis.yml"
   - $DOCKER_COMPOSE build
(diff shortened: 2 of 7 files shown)


.travis.yml
```diff
@@ -0,0 +1,36 @@
...
+cache:
+  directories:
+    - /var/lib/docker/image
+
+services:
+  - docker
+env:
+  global:
+    - DOCKER_VERSION=1.11.1-0~trusty
+    - DOCKER_COMPOSE_VERSION=1.7.1
```
sebschrader: Can we just use the default docker version instead of pinning to those specific versions?

Dockerfile
```diff
@@ -0,0 +1,64 @@
...
+
+COPY vagrant/etc/apt/sources.list /etc/apt/sources.list
+
+# Setup locale
+RUN apt-get update && \
+    apt-get install -y locales && \
+    sed -i 's/# en_US.UTF-8 UTF-8/en_US.UTF-8 UTF-8/g' /etc/locale.gen && \
+    locale-gen && \
+    dpkg-reconfigure locales && \
+    update-locale LANG=en_US.UTF-8 LC_ALL=en_US.UTF-8
```
sebschrader: Please put && on the next line/beginning of the line

Dockerfile
```diff
@@ -0,0 +1,64 @@
...
+        nodejs \
+        npm \
+        # openssh-server \
+        # postgresql \
+        # postgresql-client \
+        python-dev \
+        python-pip \
+        sqlite3 \
+        vim && \
+    ln -s /usr/bin/nodejs /usr/bin/node
```
sebschrader: See && above

Dockerfile
```diff
@@ -0,0 +1,64 @@
...
+
+# Install Bower
+RUN npm install -g bower
```
sebschrader: Can you install bower in the previous step.

I don't see the need for an additional layer here.

Dockerfile
```diff
@@ -0,0 +1,64 @@
...
+
+RUN adduser --disabled-password --gecos "Application" pycroft
+RUN mkdir -p $PROJECT_DIR/ && chown pycroft:pycroft $PROJECT_DIR
+
+# Installing the js dependencies via bower cannot be done as root
+COPY bower.json .bowerrc $PROJECT_DIR/
+RUN export BOWER_DIR=$PROJECT_DIR/web/static/libs/ \
+    && echo $BOWER_DIR \
+    && mkdir -p $BOWER_DIR \
+    && chown pycroft:pycroft $BOWER_DIR
```
sebschrader: Please remove the `$BOWER_DIR` creation

Dockerfile
```diff
@@ -0,0 +1,64 @@
...
+
+COPY . $PROJECT_DIR
+RUN chown -R pycroft:pycroft $PROJECT_DIR
+
+USER pycroft
+WORKDIR $PROJECT_DIR
+
+RUN echo "Installing js dependencies." \
+    && bower install -F \
+    && bower update -F
```
sebschrader: You can move this up and install the dependencies as root with `--allow-root`

docker-compose.yml
```diff
@@ -0,0 +1,12 @@
...
+  web:
+    extends:
+      file: docker-compose.bare.yml
+      service: web
+    volumes:
+      - ./:/pycroft
+  db:
+    extends:
+      file: docker-compose.bare.yml
+      service: web
```
sebschrader: Can you please run the PostgreSQL database on a tmpfs

Dockerfile
lukasjuhrich: I will not, since this is required if done before `COPY . /pycroft`.
The latter however is preferable, since installing bower (which has yet to be done before the `COPY`) is another layer that can be cached when re-building upon a code change not affecting some dependencies (build information, js, …).

Dockerfile
lukasjuhrich: Thanks, I will do that as well in the old `Dockerfile`, since that is where I copied it from. Below, you see I have added it “correctly”, although unintentional.

sebschrader: You can remove the locales stuff.

There is no need for locales in the container.

Dockerfile
```diff
@@ -0,0 +1,64 @@
...
+        bash \
+        curl \
+        git \
+        libpq-dev \
+        libsqlite3-dev \
+        nodejs \
+        npm \
+        # openssh-server \
+        # postgresql \
+        # postgresql-client \
```
lukasjuhrich: I should probably remove the comments.

sebschrader:
Please provide a description for this PR.

Dockerfile
lukasjuhrich: Are they even required in the old `Dockerfile`?

.gitignore
```diff
@@ -11,3 +11,6 @@ doc/_build
 .vagrant
 web/static/libs/**
 
+*.log
+.coverage
+TAGS
```
sebschrader: Where is TAGS coming from?

server_run.py
```diff
@@ -30,6 +31,17 @@ def server_run(args):
     set_scoped_session(scoped_session(sessionmaker(bind=engine),
                                       scopefunc=lambda: _request_ctx_stack.top))
 
+    if args.recreate_db:
+        if args.debug:
+            exit("Not recreating the with `--debug`, "
+                 "which usually leads to multiple restarts on file change.\n"
+                 "This would override any database change every time.")
+        print("Dropping pycroft model schema")
+        drop_db_model(engine)
```
sebschrader: This won't work if tables are renamed or removed.

The new model won't contain the old tables/indexes/etc. and therefore can't drop them.

.gitignore
mmarx: Probably from etags(1).

server_run.py
mueslo: That kind of logic also really doesn't belong in a file whose purpose it is to start the web server (or at least I fail to see the logical connection)

lukasjuhrich: @mueslo It is quite a problem that `server_start.py` doesn't work from scratch. It is a useful thing for the one, central script we use all the time to start pycroft to be able to create the schema if it doesn't exist.
If you say `run_server.py` shouldn't touch the database, then I suggest renaming it or making a different script that itself may invoke `run_server.py` (e.g. `manage.py`).

Since you always start without the schema, running the app fails. This shouldn't be the case, I think it is perfectly valid for the app to at least bootstrap the schema.

lukasjuhrich:
Looking at this branch, I move out the latest commits. They have something to do with e.g. MAC-editing (frontend feature) and nothing with the dockerfile anymore.

lukasjuhrich:
The version pinning can be removed completely. The only reason was because `version: '2'`-yml-files have been used, which required an upgrade of docker-engine and docker-compose.

By now however, these versions are shipped by default in travis.

server_run.py
lukasjuhrich: I will move it out of this branch so I can merge the pull request.

Dockerfile
lukasjuhrich: Correct, that's not really an optimization though.  It still has to be below `COPY . $PROJECT_DIR`, since that is where the bower.json lies.

PR #56, opened 2016-10-10
Get rid of hard to distinguish characters in password generation(e.g. l/1 0/O)

cvonelm:
At least in Courier, which my form was printed in, those characters are almost indistinguishable

### DIFF
--- a/pycroft/helpers/user.py
+++ b/pycroft/helpers/user.py
@@ -12,6 +12,7 @@ crypt_context = ldap_context.copy(
 
 
 def generate_password(length):
+    #without hard to distinguish characters l/1 0/O
     charset = "abcdefghijkmnopqrstuvwxyz!$%&()=.," \
               ":;-_#+23456789ABCDEFGHIJKLMNPQRSTUVWXYZ"
     return passlib.utils.generate_password(length, charset)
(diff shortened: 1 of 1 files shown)


pycroft/helpers/user.py
```diff
@@ -12,8 +12,9 @@
 
 
 def generate_password(length):
-    charset = "abcdefghijklmnopqrstuvwxyz!$%&()=.," \
-              ":;-_#+1234567890ABCDEFGHIJKLMNOPQRSTUVWXYZ"
+    #without hard to distinguish characters l/1 0/O
```
lukasjuhrich: Thanks! I will replace this comment by a docstring in a few minutes though.
By the way: [PEP8 requires](https://www.python.org/dev/peps/pep-0008/#inline-comments) exactly one space after the `#` (and at least two spaces before it if you append the comment to a line.)

toothstone:
Hi @cvonelm, unfortunately this is not the software that generated your password. We would really like to replace our current legacy system, but there are some open issues (like generating & printing the sheet with the passwort information^^). If you would like to contribute and participate in our students network, you are more than welcome. Please get in touch via mitmachen@agdsn.de :)

lukasjuhrich:
@toothstone Oh. Does that mean pycroft has been enhanced by accident?

toothstone:
@lukasjuhrich As far as I can tell, yes :)

PR #66, opened 2016-12-03
Reconstruct customly small membership intervals

lukasjuhrich:
Sometimes, instead of the regular membership fee, a smaller one has been
booked by the head of finance, to compensate for the smaller interval of
membership.  In most cases, this can be used to reconstruct the
membership intervals in higher precisions (instead of throwing an error
due to fee amount inconsistencies).

### DIFF
--- a/legacy/reconstruct_memberships.py
+++ b/legacy/reconstruct_memberships.py
@@ -41,6 +41,28 @@ def interval_count(interval_list):
     return counter
 
 
+def nextmonth(original_date):
+    return (original_date.replace(day=1) + timedelta(days=32)).replace(day=1)
+
+
+def prevmonth(original_date):
+    return (original_date.replace(day=1) - timedelta(days=1)).replace(day=1)
+
+
+def advance_some_months(original_date, months):
+    current_date = original_date
+    for _ in range(months):
+        current_date = nextmonth(current_date)
+    return current_date
+
+
+def recede_some_months(original_date, months):
+    current_date = original_date
+    for _ in range(months):
+        current_date = prevmonth(current_date)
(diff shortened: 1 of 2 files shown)


legacy/reconstruct_memberships.py
```diff
@@ -179,11 +201,61 @@ def membership_from_fees(user, semesters, n):
...
+                        "Proportional fit failed ({} vs. {} corresponds to {} months)"
+                        .format(split.amount, sem.regular_semester_fee, proportional_months))
+                num_months = int(proportional_months)
+                log.debug("The manually awarded fee corresponds to %s months", num_months)
+
+                uncropped_interval = sem_to_interval(sem)
+                intervals_below = [i for i in intervals_regular if i <= uncropped_interval]
+                intervals_above = [i for i in intervals_regular if i >= uncropped_interval]
+
+                touching_above = min(intervals_above).lower_bound.value == \
```
lukasjuhrich: I should use `begin` and `end`.

legacy/reconstruct_memberships.py
```diff
@@ -41,6 +41,28 @@ def interval_count(interval_list):
...
+    return current_date
+
+
+def recede_some_months(original_date, months):
+    current_date = original_date
+    for _ in range(months):
+        current_date = prevmonth(current_date)
+    return current_date
+
+
```
lukasjuhrich: Because I wasn't quite sure and did it the safe way. There is no special reason, you can edit it.

lukasjuhrich:
@mueslo Could you perhaps change the state of your review to resolved (however this works)?

Issue #70, opened 2016-12-04
Implement redirection mail


lukasjuhrich:
The redirection email has been forgotten to implement as part of the model and LDAP-Importer.

Besides that, the mail given in the „Email“ field of the user creation page is currently stored as the mail of the user. This is counter-intuitive, since the usual mail that is given during creation is the address for redirection.

Both mail attributes should be easily editable by the user manager.

lightning1:
I don't understand your distinction between these two types of email.
In the model only one explicit mail address is visible to me: the personal address of the user (e.g. redirection mail address). The other mail address our systems are processing is always based on the login (unix_account) and is not explicitly mentioned in the model.

This issue adresses the LDAP-Importer only, is this assumption correct?

lukasjuhrich:
@lightning1 The users mail _is_ fixed by the unix_account on our current legacy system, but it _should not_ be. Since the unix_account is fixed, e.g. typos cannot be corrected. This happened for instance with @janLo (`losinshi` instead of `losinski`).  Bottom line being, what you are wondering about is an explicit feature that doesn't impose a lot of additional logic, has been included from quite the beginnig, and should stay.

This issue addresses the LDAP-Importer concerning the correct storage of _both_ mail fields (if implemented) and the user creation concerning targeting the redirection mail and not the local mail.

lukasjuhrich:
I am not quite sure whether or not I have a misunderstanding of the purpose of the `email` attribute.
When I started working on Pycroft, I recall that I was told what I described above. However, it seems strange to me that the email regex allows for email addresses which don't belong to us. It is weird that the redirection mail address hasn't been implemented in the model for such a long time, too.

@sebschrader @janLo, could you please clarify this? I believe one of you should know this.

sebschrader:
You are right, conceptually there two different types of email attributes, namely:

1. The external email address of a user
2. The AG DSN-provided email address

We store only the first type of email address in our LDAP directory in the `email`, the second type is generated from the `uid` attribute. We could also store the second address in the LDAP directory, but this would only make sense, if we want it to be different from the `uid` attribute. I however don‘t think this is necessary. Typos in the `uid` attribute can be fixed or worked around. Both has been done in the past.

If we store the second type of address, we would need a new attribute. The [Postfix LDAP Readme](http://www.postfix.org/LDAP_README.html) has a suggestion: `maildrop`. But again, it don‘t think it‘s necessary.

lukasjuhrich:
[...]

PR #74, opened 2017-03-01
Ldap syncer

lukasjuhrich:
Left todo (most of the logic is implemented already):
- [x] Actually implement and test the object fetching (db+ldap) the exporter does in advance
- [x] Filter objects synced by who has access
- [x] Revisit folder structure (`Record` and similar should be in their own module)
- [x] Find a good way to configure the exporter
- [x] Add logging
- [x] Resolve discussion in #70 – now, `User.email` has been assumed to be the redirection mail. 
- [x] Add integration tests with fixtures e.g. testing the correct users have been filtered
- [x] Documentation!
- [ ] Look into useful triggers executing this script (user creation, password change)


pycroft/ldap_sync/action.py
```diff
@@ -0,0 +1,26 @@
+class Action(object):
+    def __init__(self, record):
+        self.record = record
+
+    def execute(self, *a, **kw):
+        raise NotImplementedError("Action subclass must implement `execute`.")
```
lukasjuhrich: Please don't mind this horseshit, this is WIP and won't end up in the code.

tests/test_ldap_sync.py
```diff
@@ -0,0 +1,96 @@
...
+
+    def test_correctly_different_record_modifies(self):
+        difference = self.record - Record(dn='test', attrs={'bar': ''})
+        self.assertIsInstance(difference, ModifyAction)
+
+    def test_record_from_orm(self):
+        assert False
+
+    def test_record_from_ldap_record(self):
+        assert False
```
lukasjuhrich: FYI: These two functions are the tests which break the build.

PR #95, opened 2017-05-11
Radius logs – UI implementation

lukasjuhrich:
TODO:
- [ ] 
- [ ]

### DIFF
--- a/hades_logs/__init__.py
+++ b/hades_logs/__init__.py
@@ -107,7 +107,8 @@ class HadesLogs:
         :param str nasportid: The port identifier (e.g. `C12`) of the
             NAS port
 
-        :returns: the result of the task
+        :returns: the result of the task (see
+                  ``get_port_auth_attempts`` in hades)
 
         :raises HadesTimeout: on timeouts, e.g. when the task takes
             too long to be executed by a worker or when the broker is
--- a/web/blueprints/helpers/log.py
+++ b/web/blueprints/helpers/log.py
@@ -1,0 +1,64 @@
+from datetime import datetime
+from functools import partial
+
+from flask import url_for
+
+from pycroft.helpers.i18n import Message
+from web.template_filters import datetime_filter
+
+def format_log_entry(entry, log_type):
+    return {
+        'created_at': datetime_filter(entry.created_at),
+        'raw_created_at': entry.created_at,
+        'user': {
+            'title': entry.author.name,
+            'href': url_for("user.user_show", user_id=entry.author.id)
+        },
+        'message': Message.from_json(entry.message).localize(),
+        'type': log_type
+    }
+
+
+format_user_log_entry = partial(format_log_entry, log_type='user')
+format_room_log_entry = partial(format_log_entry, log_type='room')
+
+
+def _parse_hades_response(response):
+    auth_date = response[3]
+    msg_parts = ["Authentication of {mac} – ".format(mac=response[2])]
+    if response[0].lower() == "auth-reject":
+        msg_parts.append(
+            "REJECTED – This should never happen!"
+            " Please contact a root of your choosing!"
+        )
+    else:
+        group = response[1]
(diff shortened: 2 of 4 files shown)


lukasjuhrich:
The current status of the radius tables is the following:
alt
@sebschrader, @fgrsnau, any input on this? I plan to add a color coding similar to the one used in the Hausuebersicht, but I am not sure if the text itself could be enhanced.

sebschrader:
The radpostauth scheme has been reworked recently. We store an array of group memberships at the time of each authentication attempt and an array of all response attribute-value pairs that were sent to switch.

This allows to greatly enhance the log, because we can show exactly what was sent to the switch.

Also, the group memberships that Hades will see (radusergroup table), will be generated by a view of Pycroft data. You probably haven't thought about how to create this view yet. But if you think about it, you could show/link the Pycroft objects (e.g. Subnets/VLANs) that represent the things Hades sees in its GroupName column.

lukasjuhrich:
I am going to merge (and _reopen_!) this pull request into develop, as I ended up doing some refactoring of the bootstrap-tables and various UI improvements resp. bugfixes. I want this to be in the current test instance ASAP.

hades_logs/parsing.py
```diff
@@ -0,0 +1,103 @@
...
+    """
+    @property
+    def accepted(self):
+        """Return whether the reply says ``"Access-Accept"``"""
+        return self.reply == "Access-Accept"
+
+    @property
+    def attributes(self):
+        """The attributes provided as a dict of lists."""
+        return attrlist_to_dict(self.raw_attributes)
```
sebschrader: It's probably better to either eagerly convert the list of attribute-value-pairs in `__new__` or cache the property with [`functools.lru_cache`](https://docs.python.org/3/library/functools.html#functools.lru_cache), because this property is needed multiple times and its always needed.

I'd go for the eager method.

hades_logs/parsing.py
```diff
@@ -0,0 +1,103 @@
...
+    @property
+    def time(self):
+        return datetime.fromtimestamp(self.timestamp)
+
+    def __bool__(self):
+        """Evaluates to :py:prop:`self.accepted`"""
+        return self.accepted
+
+    def effectively_equal(self, other):
+        relevant_attributes = set(_radius_log_entry._fields) - {'timestamp'}
```
sebschrader: Why not `self._fields`?

web/blueprints/user/log.py
```diff
@@ -0,0 +1,87 @@
...
+    from pycroft.model._all import Room, SwitchPatchPort, SwitchInterface, Switch
+    query = (
+        session.session
+        .query(SwitchInterface.name, Switch.management_ip)
+        .join(SwitchPatchPort.room)
+        .join(SwitchPatchPort.switch_interface)
+        .join(SwitchInterface.host)
+        .filter(Room.id == room.id)
+    )
+    return query.all()
```
sebschrader: `all()` returns a list and not an iterator.

web/app.py
```diff
@@ -22,6 +26,34 @@ class PycroftFlask(Flask):
         undefined=StrictUndefined
     )
 
+    def __init__(self, *a, **kw):
+        super().__init__(*a, **kw)
+        # config keys to support:
+        self.maybe_add_config_from_env([
```
sebschrader: Do we really need a proliferation of more environment variables?

vagrant/requirements.txt
```diff
@@ -22,4 +22,4 @@ psycopg2>=2.6.1,<2.7
 wrapt>=1.10.8,<1.11
 # only needed for ldap caching
 ldap3==2.1.0
-celery==4.0.2
+celery==3.1.25
```
sebschrader: You should use [`~=`](https://www.python.org/dev/peps/pep-0440/#compatible-release).

Issue #69, opened 2016-12-03
Conveniently list invalid multiple occupancies


lukasjuhrich:
Since we still need the possibility to have a list where one can conveniently spot which room has multiple inhabitants which have email access.

From the UI perspective, a little „move out“ button could be added which adds a half-open membership to „Ausgezogen“ and adds a log entry that this has been added due to removing these invalid multiple occupancies.

lukasjuhrich:
@Hannibal42 I assigned you this issue as this is a nice way to get started and fix something important.  The basic structure of this task would look like this:

- [ ] Create a new page (view function + Template, think of a) where the things will be listed – may contain dummy information at the beginning
- [ ] Build the query which gives a list of rooms and users living there
- [ ] Pass this list as a parameter to the jinja template, and find a good way to incorporate it in the page
- [ ] Add frontend tests

PR #100, opened 2017-06-19
2 users1 room

Hannibal42:
See issue #69

### DIFF
--- a/tests/frontend/test_facilities.py
+++ b/tests/frontend/test_facilities.py
@@ -52,7 +52,7 @@ class Test_010_Building(FrontendDataTestBase):
                 self.building.id, self.room.level),
             "facilities/rooms.html")
 
-    def test_0090_overcrowded_rooms(self):
+    def test_overcrowded_rooms(self):
         self.assert_template_get_request(
             "/facilities/overcrowded",
             "facilities/room_overcrowded.html")
--- a/tests/frontend/test_facilities.py
+++ b/tests/frontend/test_facilities.py
@@ -51,3 +51,8 @@ class Test_010_Building(FrontendDataTestBase):
             "/facilities/buildings/{}/levels/{}/rooms/".format(
                 self.building.id, self.room.level),
             "facilities/rooms.html")
+
+    def test_0090_overcrowded_rooms(self):
+        self.assert_template_get_request(
+            "/facilities/overcrowded",
+            "facilities/room_overcrowded.html")
--- a/web/blueprints/facilities/__init__.py
+++ b/web/blueprints/facilities/__init__.py
@@ -26,6 +26,8 @@ from web.blueprints.facilities.forms import (
 from web.blueprints.helpers.user import user_button
 from web.blueprints.navigation import BlueprintNavigation
 from web.template_filters import datetime_filter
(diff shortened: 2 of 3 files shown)


tests/frontend/test_facilities.py
```diff
@@ -51,3 +51,8 @@ def test_0080_building_level_rooms(self):
             "/facilities/buildings/{}/levels/{}/rooms/".format(
                 self.building.id, self.room.level),
             "facilities/rooms.html")
+
+    def test_0090_overcrowded_rooms(self):
```
lukasjuhrich: Great that you provided a frontend test!

Minor detail, please remove the number from your new test case.
I know a lot of tests contain these, but including a number to enforce an execution order is pointless and unidiomatic.
Most of the tests have been left that way, but don't let that confuse you.

Hannibal42:
Done.

sebschrader:
Some rooms are allowed to be occupied by more than one person. Such rooms are excluded by the current system used at Wundtstraße as long as there were not more than two inhabitants.

web/blueprints/facilities/__init__.py
```diff
@@ -40,6 +42,36 @@ def root():
...
     return render_template('facilities/site_overview.html')
 
+@bp.route('/overcrowded')
+@nav.navigate(u"Two users one room")
+def overcrowded():
+    return render_template("facilities/room_overcrowded.html")
+
+@bp.route('/overcrowded/json')
+def overcrowded_json():
+    query = session.session.query(User.room_id, func.array_agg(User.id), func.array_agg(User.name))\
```
lukasjuhrich: Just querying for the Room ID is not sufficient, we need the full ORM object in order to display a useful string representation. However, something like `Room.q.join(User).group_by(Room.id).having(func.count(User.account_id) > 1).add_columns(func.array_agg(User.id)).all()
` does not work for a curious error („Unhashable type: List“)

lukasjuhrich:
@Hannibal42 Do you have capacities to finish this PR?

Perhaps @sebschrader has some ideas how we could efficiently fetch all ORM user objects without doing something like `User.q.get(i)` in a for-loop over the last query's result.

sebschrader:
The SQL query you suggested doesn't male sense. You can't group by some columns A and then access columns other columns than A without using aggregate functions on these columns.

I'll try to take a look at lit later. Probably a window function will solve the problem.

Hannibal42:
I can have a look in the next few days, hopefully at saturday. But i first have to setup pycroft on my new system.

sebschrader:
A window function is not necessary. The following self semi-join with the result of grouping rooms by the people that live three should do the trick.

```sql
SELECT r1.*
  FROM room r1 JOIN (
    SELECT r2.id
    FROM room r2
    JOIN "user" u
    ON u.room_id = r2.id
    GROUP BY r2.id
    HAVING COUNT(u.id) > 1
  ) r2
  ON r1.id = r2.id;
```

If I find the time, I try to translate the query to SQLAlchemy ORM.

ibot3:
Implemented with https://github.com/agdsn/pycroft/pull/276.

PR #112, opened 2017-09-14
Traffic fixes

### LINKED ISSUE #111: TrafficVolume is missing a user_id foreign key
The `TrafficVolume` model class has no explicit relationship with a user, only transitively through `IP` to `Interface` to `Host` to `User`.

This is not good, because IP ownerships change.

Georgeto:
Closes #111

### DIFF
--- a/bower.json
+++ b/bower.json
@@ -12,6 +12,7 @@
     "bootstrap-table": "~1.11.1",
     "underscore": "~1.7.0",
     "d3": "3.x",
-    "dcjs": "~2.0"
+    "dcjs": "~2.0",
+    "nvd3": "~1.8.6"
   }
 }
--- a/pycroft/lib/user.py
+++ b/pycroft/lib/user.py
@@ -314,19 +314,13 @@ def traffic_events_expr():
         literal("credit").label('type')]),
 
         select([(-TrafficVolume.amount).label('amount'),
-                Host.owner_id.label('user_id'),
+                TrafficVolume.user_id,
                 TrafficVolume.timestamp,
                 literal("debit").label('type')]
-               ).select_from(
-            TrafficVolume.__table__.join(
-                IP.__table__
-            ).join(
-                Interface.__table__
-            ).join(
-                Host.__table__)),
+               ),
 
         select([TrafficBalance.amount,
-                TrafficBalance.user_id.label('user_id'),
+                TrafficBalance.user_id,
                 TrafficBalance.timestamp,
                 literal("balance").label('type')]  # ).select_from(
                # .__table__.outerjoin(TrafficBalance)
--- a/pycroft/lib/user.py
+++ b/pycroft/lib/user.py
@@ -330,7 +330,7 @@ def traffic_events_expr():
     return events
 
 
-def traffic_balance_expr():
+def traffic_balance_expr(until=func.now()):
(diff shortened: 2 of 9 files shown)


sebschrader:
The problem with very frequent traffic balance calculations is that there can be only one TrafficBalance per user. You lose all history if you create a new balance. History should be kept for a little while (few days) and therefore we need to bind the traffic to the user as early as possible.

mueslo:
You don't lose any history until IP objects start getting deleted. Just because a new balance calculated later in time doesn't mean you can't reconstruct the balance prior to that timestamp.

sebschrader:
But only if it's guaranteed that a user's IP doesn't change, which it isn't.

sebschrader:
I would like to get rid of a the foreign key relationship to the IP object and store the IP address as value directly.

Georgeto:
The previous traffic graph had quite a few issues (not all visible in this screenshot).
screenshot_20170923_001527

Therefore I created a new one and also added a credits graph.
screenshot_20170929_130029

mueslo:
What was your issue with the previous traffic graph except missing binary prefixes? 

But more importantly, your balance calculations are incorrect. The balance timestamp could be anywhere in the past, i. e. also within any of the timespan windows.

Georgeto:
> What was your issue with the previous traffic graph except missing binary prefixes?
1. Displayed timespan is not the selected one, but instead derived from data.

2. Visual glitches
screenshot_20170923_001546

3. No proper resizing, when the available size changes after initial construction.
scalecomparison

4. Implemented using of a low level api, and therefore harder to understand and maintain.
5. I like the design of the new graphs more, but I guess thats up to preference.


> But more importantly, your balance calculations are incorrect. The balance timestamp could be anywhere in the past, i. e. also within any of the timespan windows.
I don't think my calculations are wrong. The balance value returned by the json_trafficdata endpoint is the sum of all traffic data points (balance, volume, credits) up until to the first data point returned in debits/credits.

mueslo:
> The balance value returned by the json_trafficdata endpoint is the sum of all traffic data points (balance, volume, credits) up until to the first data point returned in debits/credits.

There is only one balance entry. As I wrote, this could be e.g. from 2 days ago. If you sum events up until 7 days ago, you will not get the correct balance. The json_trafficdata endpoint uses the lib-function you edited. It didn't have an 'until' parameter for good reason, because all that is given about the traffic balance entry is that it sometime in the past.

Georgeto:
Mhm ok my bad, I will fix this.

lukasjuhrich:
@Georgeto what's the progress here? What big a problem do the issues mentioned by @mueslo create?

Georgeto:
I decided to rework the `json_trafficdata` API and do the data aggregation server side.
So I wrote an according database query, but then stopped there.
Hopefully I will be able to allocate some time for finishing this soon.

lukasjuhrich:
@Georgeto FYI, I started to cherry-pick the only leftover commit that's not merged yet.

Apart from that, this is what's left to do:
- [ ] Cherry-pick the traffic-graph improvements
- [ ] Fix traffic calculations regarding the balance entry, check everything is done as indented
- [ ] Remove FKey constraint TrafficVolume→IP (how hard is that? What implications does it have? Tackle the issue or postpone?)

PR #115, opened 2017-09-29
Minor user log improvements

Georgeto:
As a result of the refactorings in 3403eaee4d99be27bc463279583d356fc0b6caaa the now deleted javascript could no longer fulfill its purpose. As I do not think we need ajax powered user log entry creation, I have decided to simply delete it.


lukasjuhrich:
@georgeto you can merge this branch by rebasing on top of develop (so everything stays linear and we can spare the merge commit), as I can't use git at the moment.

PR #117, opened 2017-10-27
Frontend: Unlocking-Button resolves #2

### LINKED ISSUE #2: Frontend: Unlocking should work the same way as locking
> ### Redmine metadata
> 
> | Attribute | Value |
> | --- | --- |
> | done_ratio | 0 |
> | status | New |
> | start_date | 2016/01/30 |
> | created_on | 2016/01/30 23:14:18 +0100 |
> | author | @lukasjuhrich |
> | priority | Normal |
> | updated_on | 2016/09/18 20:00:37 +0200 |
> | tracker | Bug |
> | asigned_to | None |

---

To unlock a user who has been locked can be done via terminating the membership in the `Gesperrt` Group.

Since locking a User happens using the „common actions“ buttons in the overview section, it makes sense to change the “Sperren” button to “Entsperren” (while adding actual unlocking functionality to said button) if a user is member of the `Gesperrt` group.
## 

_This issue has been automatically imported from redmine on 2016-09-28 19:06:55.163100_

marcelb98:


web/blueprints/user/__init__.py
```diff
@@ -151,6 +151,11 @@ def user_show(user_id):
         'inverted': True,
         'saldo': balance,
     }
+    lock_unlock="lock" # lock_unlock is "lock" if user isn't locked --> show button to lock user | if user is locked, lock_unlock is the membership-ID
+    for membership in memberships_active.all():
+        if membership.group.name == "Gesperrt":
+            lock_unlock=membership.id
+            break
```
lukasjuhrich: Instead of hard-coding the string „Gesperrt“ (which the according group does not need to have!), you should reference the group that is given in the config as done [here](https://github.com/agdsn/pycroft/blob/de8ad9742904ef7c51ed9a8accb55c9ded40bd52/pycroft/lib/user.py#L428): the corresponding group would be `config.violation_group`.

Although this works, we can end the membership even simpler: The library function `remove_member_of` in `lib.user` only needs to know the group, not the specific membership or its id. Thus, it would be more fitting to:

- [ ] create a new viewpoint `/unblock` (see the `suspend` function for comparison)
- [ ] create a new library function `def unblock(user, processor)` (see the `suspend`) that calls `remove_member_of(user=user, group=config.violation_group, processor=processor)`
- [ ] Use this new endpoint in the template.

web/templates/user/_user_show_basedata.html
```diff
@@ -69,7 +69,11 @@
     {% if current_user is privileged_for('user_change') %}
     <div class="col-sm-6 col-md-5 col-lg-4">
         <div class="user"><a class="btn btn-info" href="{{ url_for(".move", user_id=user.id) }}">Umziehen</a></div>
-        <div class="user"><a class="btn btn-info" href="{{ url_for(".suspend", user_id=user.id) }}">Sperren</a></div>
+        {% if lock_unlock == "lock" %}
+          <div class="user"><a class="btn btn-info" href="{{ url_for(".suspend", user_id=user.id) }}">Sperren</a></div>
+        {% else %}
+          <div class="user"><a class="btn btn-info" href="{{ url_for(".end_membership", user_id=user.id, membership_id=lock_unlock) }}">Entsperren</a></div>
+        {% endif %}
```
lukasjuhrich: as mentioned above, use `{% if is_locked %}` and call the new endpoint `".unblock"`

lukasjuhrich:
@marcelb98 I extended what you did by a frontend test case (which lead to various other stuff such as bumping dependencies).

pycroft/lib/user.py
```diff
@@ -433,6 +438,24 @@ def suspend(user, reason, processor, during=None):
...
+    """Unblocks a user.
+
+    This removes his membership of the ``config.violation`` group.
+
+    :param User user: The user to be unblocked.
+    :param User processor: The admin who unblocked the user.
+
+    :return: The unblocked user.
+    """
+    remove_member_of(user=user, group=config.violation_group,
```
lukasjuhrich: This is a bug, without giving `during=closedopen(when, None)` this removes every membership in the group retroactively instead of terminating.

PR #118, opened 2017-11-02
WIP: Remove away

### LINKED ISSUE #90: Remove `away` group
In times of older finance regulations, there was a reduced membership consisting only of email.
It is correct that users will be added to these groups by the importer if a reduced fee has been booked.

However, this group shouldn't be used in the _current_ library anymore.

Thus, at least `lib.user.move_out_temporarily` should be rewritten in order to not add an `away` membership, but to add a `closedopen()` membership in `member` _after_ the given absence period ended. (Refs #46)

To be done:
- remove `away` group from config and library

lukasjuhrich:
Fixes #90


pycroft/lib/user.py
```diff
@@ -467,75 +503,41 @@ def move_out(user, comment, processor, when):
...
     """
-    for group in (config.member_group, config.network_access_group):
+    if when > datetime.now():
+        raise NotImplementedError("Moving out in the future is not supported yet.")
+
+    for group in {config.member_group, config.network_access_group}:
         remove_member_of(user, group, processor, closedopen(when, None))
 
+    num = 0
+    for num, h in enumerate(chain(user.user_hosts, user.server_hosts)):
```
lukasjuhrich: Wait, this is stupid. Note to self: Use `enumerate(hosts, start=1)`

lukasjuhrich:
TBH I'm a bit overwhelmed by all the edge cases that can occur (considering `move_back_in`).

For instance, you can run the same `move_back_in` POST query multiple times and get a change that is idempotent in the room and in the memberships, but a second host is added with the same MAC address.

In this case it's quite obvious to restrict the action for users with `not user.member_of(config_group)`, but I am unsure whether to place this requirement in the view function only or in the library function as well. If the latter, how coarsely or densely should that be tested?

Also, what if the user is not in the member group, but has a room? Or the converse? how much of such Edge cases should be supported? (inb4 as little as possible, as much as necessary)

@sebschrader @lightning1 @MarauderXtreme some comments would be appreciated.

sebschrader:
Are your questions/concern about the idempotency of the lib functions still valid, since you've already merged this pull requests?

lukasjuhrich:
The questions still persist. It is not too relevant at the moment, because the only function using `move_back_in` is the view function in `web` which tests for `user.is_member(config.member_group)`, but if the `lib` will be exposed e.g. as a REST API, the function may be called under more weird circumstances.

Issue #98, opened 2017-05-21
Phrase “mail access” like “access to services”


lukasjuhrich:
Although at the present time, the most important thing about having an LDAP account is the email access, I find it more appropriate to refer to this as “access to services”. Firstly, having an LDAP account also gives access to the webhosting over at [users.agdsn.de](https://users.agdsn.de/), and secondly, this might be expanded in the future, since Pycroft aims to provide a central member's LDAP.

Therefore, I propose to change the descriptions referring to “mail access” to “access to services”, and renaming the property `mail` to `services`.

Perhaps this should be done before the first deployment as this would imply a small, but significant structural change (concerning the static contents in the `config` table).

sebschrader:
Specific properties for specific services would be the best thing.

lukasjuhrich:
@sebschrader I disagree. This would be an unnecessary refactoring burden and there is no real disadvantage in bundling the access to one property.

sebschrader:
The whole point (or one important) of properties is to have a very granular mechanism instead of a tight coupling of our policies with our code.

Besides, none of the Pycroft code, should actually need to know about the properties our mail system or our public login server. They verify their properties (through an eventual LDAP export) and pycroft verifies its properties.

lukasjuhrich:
I am not sure I agree this is worth the effort. We have never handled our services separately, either you had a unix account and thus access to `userlogin`, the userhosting, your mailbox, and the userdb, or you didn't. Currently however, naming it `mail` is misleading and wrong, as it represents more than just access to mail, and without reimplementing a whole lot of logic, renaming it would be the more rational option.

sebschrader:
If adding new groups and properties is such a big effort, then clearly something is very wrong.

There are some groups that are special to Pycroft, i.e. the ones referenced by the `Config` model class.
Changing or adding those is clearly an effort, but this by design and ok. They are special to Pycroft, because Pycroft automatically manages group memberships in these groups. Other, manually managed groups, should however always be possible.

Adding new properties on the other hand, should never be a problem. If other services want to leverage Pycroft's property system, the should be able to do so. We can always add properties to the `member_group` for example and Pycroft won't care.

lukasjuhrich:
I understand your point now.

I found the following occurrences of “hard-coding” the mail attribute:
- `lib.user`, in `status()` and `status_query()` (the logic providing the letter-icon in the status buttons)
- `model.property`, in the hard-coded property categories. These only provide descriptions.

[...]

PR #125, opened 2017-12-04
Mail vs ldap (WIP)

lukasjuhrich:


lukasjuhrich:
(to be) referencing #98 and #7

Issue #119, opened 2017-11-05
`move_in` does not give initial credit


lukasjuhrich:
The `move_in` routine should add an initial `TrafficBalance` of a configurable amount (usually 21GiB).

lukasjuhrich:
Combined with this, we need to create an interface for finance managers to book the monthly contributions.

lukasjuhrich:
Fixed by c4d607b2b9b6cea92ac32a5253593fc8a756c406

Issue #89, opened 2017-04-07
Add Users to traffic_group when moving in or moving


lukasjuhrich:
Adding users to traffic groups is only done in the initial legacy import. At least in `pycroft.lib.move_in`, they should be added to the respective traffic_group.

However, at this point it is still unclear which traffic group should be chosen for which users. Although there is some sort of “default traffic group”, which is the one all of our members will end up in, there are others. For instance, for guests of the IGH, different traffic regulations hold, so they need to be assigned to a different `traffic_group` automatically.

Of course, switching traffic groups manually is always possible for user managers, as this requires just a change of memberships.

lukasjuhrich:
@mueslo is this WIP? When reworking docstrings in `lib.user`, I noticed that `move_in` does not even add initial credit

PR #135, opened 2017-12-11
WIP: Credit granting && traffic group association

lukasjuhrich:
Left todo:

- [x] Unittests for `determine_traffic_group` and `setup_traffic_group`
- [x] Make `default_traffic_group` a relationship and use `default_traffic_group_id`
- [x] Fix the latter in the importer
- [x] Finish && test credit granting
- [ ] ~Add UI selection of custom traffic group (perhaps separate issue because not using the default group is really an edge case)~ see #137 

Refs #119 #89

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -43,7 +43,7 @@ dist/
 downloads/
 eggs/
 .eggs/
-lib/
+/lib/
 lib64/
 parts/
 sdist/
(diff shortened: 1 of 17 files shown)


lukasjuhrich:
@Georgeto although I'm not 100% sure what you do at the moment, keep an eye on my changes. For instance, I refactored some functions out of `user.lib`.

Issue #94, opened 2017-05-08
Create Traffic Group: Broken interval field


toothstone:
Traffic groups require an interval, but the input field allows only dates, and so the form validation fails on submission.

Steps to reproduce:
1. Go to /properties/traffic_group/create
2. Fill name, amount and cap.
3. Try to enter a meaningful interval ⇒ converted to current date, date picker shown
4. Submit

lukasjuhrich:
I just realized my fix kind of broke it, because the field values now default to an empty string instead of `0 mons … 0 secs`.

PR #140, opened 2018-01-05
Create trafficgroup

marcelb98:
- fixed issue #94 
- new field-type for interval input
- field has picker to assist user inputting a correct formatted interval
- field has validation

Furthermore, I've changed some fields in the form from `TextField` to `IntegerField`.

I've also added "Initial Credit" in the form and on the trafficgroups overview site.


lukasjuhrich:
Thanks again for the quick fix! Because my tests and changes depended on the `factory_boy` branch which (now) is an ancestor of yours, I will not add a review for these specific changes. An extensive commentary has been sent via E-mail.

PR #144, opened 2018-01-20
Automatic schema upgrading

### LINKED ISSUE #93: Add schema version control
Using a tool like [alembic](http://alembic.zzzcomputing.com/en/latest/tutorial.html), the schema has to be under version control at the time of the first deployment.

In the course of the above, we should provide an easy way of bootstrapping the pycroft database (ideally in the entrypoint of the docker container) so that the user does not have to run manual commands to have a working pycroft instance up-and-running.

lukasjuhrich:
Fixes #93 

- [x] Check the state of the schema before running
- [x] Add stamping of the `head` revision to the importer
- [x] Automatically Call alembic's upgrading routine
- [x] Add routine to create the model and stamp `head`


lukasjuhrich:
We need to talk project structure here for a minute.

I just threw the basic blocks into `server_run.py` because that's where they're needed. However, we need this code at two other locations:

- the importer will want to call `AlembicState.stamp_head()`
- the desired setup-routine that creates the db model and then stamps as well.

The following are solutions I dislike:
- Adding it to `pycroft.model` – the `alembic` dependency is on the management level, and does not concern the model logic.
- Just throwing it into the top directory is messy and not the right place
- Adding it to `alembic/` does not seem right because it's a system generated directory and nobody would suspect anything besides migration scripts from there.
- Adding it to `helpers` is not right because there is only code specific to development and e.g. our travis setup, which means there should absolutely not be code that is a dependency for other code.

That being said, I realize that `server_run.py` is not a part of the pycroft package, and neither is the importer, so probably some location in neither should do the trick.

sebschrader:
Well, `server_run.py` should probably become part of a package instead. With setuptools, it can still be used as a console script through the [entry points](https://setuptools.readthedocs.io/en/latest/setuptools.html#automatic-script-creation) mechanism. A proper `setup.py`, would also be a big improvement anyway.

The whole package name situation BTW has been problem since, well, forever. There should be no `ldap_sync`, `legacy`, `legacy_gerok`, `web` top level packages. They should become sub-packages of `pycroft`.

It was once intended to the project code into `pycroft_web` and `pycroft_lib`, but honestly, I don't see any avenue for Pycroft code reuse in other projects at this point, so I'm not advocating for it. Other projects should preferably use (yet not existing) APIs instead of pycroft code.

lukasjuhrich:
> With setuptools, it can still be used as a console script through the [entry points](https://setuptools.readthedocs.io/en/latest/setuptools.html#automatic-script-creation) mechanism. A proper `setup.py`, would also be a big improvement anyway.

I know, because I already [implemented it](https://github.com/agdsn/pycroft/blob/develop/setup.py#L21). :P

sebschrader:
> I know, because I already implemented it. :P

Well all the better!

lukasjuhrich:
@sebschrader the solution I chose was to add a `scripts` package and add `scripts.server_run` retaining the `pycroft = scripts.server_run.main` script mapping.

PR #145, opened 2018-01-21
WIP: Views and rules

### LINKED ISSUE #111: TrafficVolume is missing a user_id foreign key

### LINKED ISSUE #122: Link pycroft to pmacct
This issue depends on the developments in the `views_and_rules` branch, where a custom compiler has been introduced that understands rules. A WIP commit introduced the inbound and outbound tables and sketched creation of the rules, however, the rule's query has not been implemented yet in the rule creation.

lukasjuhrich:
Fixes #111, about to fix #122 

Left todo:

- [x] Compiler bug:  `DropFunction` misses `cascade` attribute
- [x] Compiler bug: `DropFunction` quotes the whole function identifier including parameters, not just the name: `DROP FUNCTION IF EXISTS "account_is_type(integer, account_type)"` is issued, `DROP FUNCTION IF EXISTS "account_is_type"(integer, account_type)` would be correct.
- [x] Include creation query for pmacct ghost-table rule
- [x] Test pmacct table rule

### DIFF
--- a/pycroft/lib/user.py
+++ b/pycroft/lib/user.py
@@ -364,19 +364,13 @@ def traffic_events_expr():
         literal("credit").label('type')]),
 
         select([(-TrafficVolume.amount).label('amount'),
-                Host.owner_id.label('user_id'),
+                TrafficVolume.user_id,
                 TrafficVolume.timestamp,
                 literal("debit").label('type')]
-               ).select_from(
-            TrafficVolume.__table__.join(
-                IP.__table__
-            ).join(
-                Interface.__table__
-            ).join(
-                Host.__table__)),
+               ),
 
         select([TrafficBalance.amount,
-                TrafficBalance.user_id.label('user_id'),
+                TrafficBalance.user_id,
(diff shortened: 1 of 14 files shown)


lukasjuhrich:
Concerning the quoting issue, I would adhere to the quickfix of not applying any quoting at all. The situation where it's needed is an edge case, and even then it's simpler to just specify the quoting manually, like `ddl.Function('"account is bad"(integer)', "SELECT 'a'")`.

sebschrader:
A better but still simple fix would be to split the `name` argument of `Function` into `name` and a list of arguments and apply quoting to the function name only. The list of arguments should be treated as an opaque string and not be quoted, because the function arguments can be quite complex. They have a name, a type, an optional default value.

lukasjuhrich:
This branch is good to go, after the tests pass I'll merge.

PR #163, opened 2018-03-05
User Datasheet generation

strebbin:
Allows to download an Datasheet after generating a new User


AUTHORS
```diff
@@ -22,3 +22,4 @@ Philipp Horn <unicorn@wh2.tu-dresden.de>
 Sebastian Schrader <shreyder@wh2.tu-dresden.de>
 Till Smejkal <till.smejkal@mailbox.tu-dresden.de>
 Tim Kluge <timklge@wh2.tu-dresden.de>
+Stefan Haller <stefanhaller@wh2.tu-dresden.de>
```
lukasjuhrich: Please adhere to the alphabetical order (I'm sorry @strebbin, it appears you inherited some issues ;) )

pycroft/helpers/printing/__init__.py
```diff
@@ -0,0 +1,81 @@
...
+
+    im = Image(ASSETS_LOGO_FILENAME, 5 * cm, 5 * cm)
+    story.append(im)
+    story.append(HRFlowable(width="100%", thickness=3, color=black, spaceBefore = 0.8*cm, spaceAfter=0.8*cm))
+    story.append(Paragraph('''Welcome as a member of the AG DSN, {}!'''.format(user.name),style['BodyText']))
+    story.append(Paragraph('''We are proud to announce that your network access has been activated. If you encounter any problems, drop us a mail or visit us during our office hours. You can find contact information at the bottom of this page.''', style['BodyText']))
+    story.append(Paragraph('''Please make sure to pay your membership contribution in time. You can find further details on our web page.''', style['Bold']))
+    story.append(Paragraph('''Wishing you all the best,''', style['BodyText']))
+    story.append(Paragraph('''Your AG DSN''', style['BodyText']))
+    story.append(HRFlowable(width="100%", thickness=3, color=black, spaceBefore=0.8*cm, spaceAfter=0.8*cm))
```
lukasjuhrich: although we don't enforce the 80 chars, please adhere to the soft limit of 90 and break appropriately using one of either indentation styles:
```python
some_function_call(
    "fooo bar bar bar bar bar bar bar bar"
    " baz shizzle {user} {pw}"
    .format(user="you", pw="secret"),
    msg="don't forget the trailing comma!",
)

some_other_long_thing_to_bear_in_mind(map(my_list,
                                          some_function),
                                      bracket_ends_on_this_line)
```
Note that you can wrap strings rather easily as seen above, so multiline strings using `"""` are not necessary.

pycroft/helpers/printing/__init__.py
```diff
@@ -0,0 +1,81 @@
...
+    stylesheet.add(ParagraphStyle(name='BodyText',
+                                  parent=stylesheet['Normal'],
+                                  spaceBefore=14)
+                   )
+    stylesheet.add(ParagraphStyle(name='Bold',
+                                  parent=stylesheet['BodyText'],
+                                  fontName = "Helvetica-Bold")
+                   )
+
+    return stylesheet
```
lukasjuhrich: This is missing a newline before EOF

pycroft/helpers/printing/__init__.py
```diff
@@ -0,0 +1,81 @@
...
+
+def getStyleSheet():
+    """Returns a stylesheet object"""
+    stylesheet = StyleSheet1()
+
+    stylesheet.add(ParagraphStyle(name='Normal',
+                                  fontName="Helvetica",
+                                  fontSize=10,
+                                  leading=12)
+                   )
```
lukasjuhrich: Please be consistent when indenting in this version and don't put the closing bracket on a separate line. It's not technically PEP8 but the more symmetric choice, either opening and closing bracket are inwards-isolated or none.

pycroft/lib/user.py
```diff
@@ -93,6 +93,12 @@ def setup_ipv4_networking(host):
                     subnet=subnet)
         session.session.add(new_ip)
 
+@with_transaction
```
lukasjuhrich: Two empty lines surrounding a top-level function

tests/lib/test_user.py
```diff
@@ -181,7 +181,7 @@ def test_0030_move_out(self):
         test_mac = "12:11:11:11:11:11"
         traffic_group = user.TrafficGroup.q.first()
 
-        new_user = UserHelper.move_in(
+        new_user, plain_password = UserHelper.move_in(
```
lukasjuhrich: Just a note: if you don't actually use that variable, you can replace it with an underscore: `new_user, _ = UserHelper.move_in(…)`

web/blueprints/user/__init__.py
```diff
@@ -377,17 +395,22 @@ def create():
...
             )
+
+            pdf_data = b64encode(generate_user_sheet(new_user, plain_password)).decode('ascii')
+            pdf_storage = WebStorage(data=pdf_data, expiry=datetime.utcnow() + timedelta(minutes=15))
+            session.session.add(pdf_storage)
             session.session.commit()
 
-            flash(u'Benutzer angelegt', 'success')
+            flasksession['user_sheet'] = pdf_storage.id
+            flash(Markup(u'Benutzer angelegt. <a href="{}">Nutzerdatenblatt</a> verfügbar!'.format(url_for('.user_sheet'))), 'success')
```
lukasjuhrich: View functions should be a simple, replacable frontend (always think of the repetition that would occur when introducing the same functionality in a REST API).
Instead of doing things here like passing the binary data of the user sheet somewhere, this (including adding a new WebStorage object to the session) can be handled in a library function that returns the resulting storage object. The helper could be extended by a function that encapsulates the whole `b64-> ascii` encoding procedure.

web/blueprints/user/__init__.py
```diff
@@ -630,6 +653,24 @@ def list_users_by_traffic_group(traffic_group_id):
                            users=user_list)
 
 
+@bp.route('/<int:user_id>/reset_password', methods=['GET', 'POST'])
+@access.require('user_change')
+def reset_password(user_id):
+    form = UserResetPasswordForm()
```
lukasjuhrich: Why wuold a form be needed? Just acting upon GET requests by just doing that would be sufficient.
Implementing some confirmation mechanism can just as well be handled by using a GET parameter `confirm`.

pycroft/lib/user.py
```diff
@@ -93,6 +93,12 @@ def setup_ipv4_networking(host):
...
+def reset_password(myUser):
+    plain_password = user_helper.generate_password(12)
+    myUser.password = plain_password
+    return plain_password
```
lukasjuhrich: This is missing the second most important feature: Adding a log entry saying the password has been resetted (and thus the function is missing something like a `processor` parameter)

vagrant/requirements.txt
```diff
@@ -28,3 +28,5 @@ factory-boy~=2.9.2
 alembic~=0.9.6
 # decimal serialization
 simplejson~=3.11.1
+#usersheet generation
+reportlab~=3.4.0
```
lukasjuhrich: +1 for commenting and using compatible release matching. Nice!

AUTHORS
fgrsnau: To be honest, I did expect to contribute more code to pycroft than such a little unfinished change. :smile:

pycroft/lib/user.py
mueslo: lukasjuhrich > [HoundCI](https://houndci.com/)

web/blueprints/user/__init__.py
sebschrader: Don't use german strings please.

web/blueprints/user/__init__.py
```diff
@@ -630,6 +653,24 @@ def list_users_by_traffic_group(traffic_group_id):
...
+        pdf_storage = WebStorage(data=pdf_data, expiry=datetime.utcnow() + timedelta(minutes=5))
+        session.session.add(pdf_storage)
+        session.session.commit()
+
+        flasksession['user_sheet'] = pdf_storage.id
+        flash(Markup(u'Passwort erfolgreich zurückgesetzt. <a href="{}">Nutzerdatenblatt</a> verfügbar!'.format(url_for('.user_sheet'))), 'success')
+        return redirect(url_for('.user_show', user_id=user_id))
+    return render_template('user/user_reset_password.html', form=form, user_id=user_id)
+
+
```
sebschrader: German strings again.

toothstone:
> The data sheet has no localization yet. But we need that definitely.

Well currently we are quite fine without any. Also, how do you plan to detect the language needed?

sebschrader:
>   Well currently we are quite fine without any.

Yeah, well currently the data sheet mainly uses words like `IP-Adresse`, which are almost the same in German and English.

Also there are no paragraphs in the old sheet, just attributes and values. This sheet contains actual paragraphs.

> Also, how do you plan to detect the language needed?

I wouldn't detect it, I would select it. If somebody submitted an English application form, he probably wants an English data sheet back.

It would generally be useful to store the user's preferred locale with his profile in out database, so we should add a selection to the user creation form and store it in the database. This information is useful in various other places too, e.g. E-mails we want to send out, default locale when logging with SIPA, etc.

strebbin:
> Also there are no paragraphs in the old sheet, just attributes and values. This sheet contains actual paragraphs.

As @FriwiDev Noted in Issue #8 the 'old' datasheet has changed. It currently has the same paragraphs with descriptions as the version in this PR. (I took the texts from there.)

web/blueprints/user/__init__.py
strebbin: I hope @fgrsnau can answer this, as i have not changed this part from his old commit.

web/blueprints/user/__init__.py
```diff
@@ -653,7 +653,7 @@ def reset_password(user_id):
     form = UserResetPasswordForm()
     myUser = User.q.get(user_id)
     if form.validate_on_submit():
-        plain_password = lib.user.reset_password(myUser)
+        plain_password = lib.user.reset_password(myUser, current_user)
```
lukasjuhrich: Please be more descriptive and add a keyword argument (`processor=current_user`). At first sight you just pass two users whose role as arguments is not easily distinguishable.

strebbin: good point.

lukasjuhrich:
@strebbin kudos to your fast response time.

sebschrader:
> As @FriwiDev Noted in Issue #8 the 'old' datasheet has changed. It currently has the same paragraphs with descriptions as the version in this PR. (I took the texts from there.)

That may very well be. I'm still not convinced, that we should forego localization.

> Well currently we are quite fine without any.

We're just too lazy and the users don't care about the sheet, because the information is mostly irrelevant to them.

My point still stands, that it is very useful to know the preferred language of the user for various other reasons. This is probably just the first time this came up in the context Pycroft.

Also, localization support is not very hard. Look at SIPA, if you want an example.

But this time, please keep the default language in the code English, not like in SIPA, where the default strings in the code are German and English is a translation. Tools, e.g `poedit`, don't cope with this very well, because they assume English is the default.

sebschrader:
Just to be clear in case it isn't obvious. I'm not talking about localizing Pycroft's user interface, just the strings that our members get to see.

strebbin:
@sebschrader I agree with you, that it would be nice to have an localized datasheet. But I think it is not necessary for now. To have an english-only datasheet is no regression to the current situation so we don't need this for our Bare Minimum. There are many projects which wait for the Bare-Minimum-completion of Pycroft. 

Till the end if this month I have no time to work on the localization and i don't want to delay the required feature of having an User-Datasheet for the Nice-to-have-Feature to have it localized. For this reason i would like to split this topic from that PR.

toothstone:
> Also there are no paragraphs in the old sheet, just attributes and values. This sheet contains actual paragraphs.

Sorry, but you are quite out of touch ;)
[This](https://user-images.githubusercontent.com/12697767/31180393-ce11a61e-a91e-11e7-8bbd-98033ca6cf1a.png) is more or less the currently printed data sheet.

> I wouldn't detect it, I would select it. If somebody submitted an English application form, he probably wants an English data sheet back.
Fair enough.

Anyhow, imho localizing this is by far not an urgent issue and should not postpone this merge. Given your further thoughts on the topic, a seperate issue seems more appropriate.

sebschrader:
> Sorry, but you are quite out of touch ;)
> This is more or less the currently printed data sheet.

Has already been mentioned. I've actually seen this sheet before, I just forgot about it.

> Anyhow, imho localizing this is by far not an urgent issue and should not postpone this merge. Given your further thoughts on the topic, a seperate issue seems more appropriate.

If you insist.

lukasjuhrich:
What's missing are the following test cases (integration is enough, we don't need to test every part of the WebStorage logic IMO)

- [ ] Test that calling `user/<user_id>/datasheet` returns a pdf with the correct content-disposition
- [ ] Test that visiting the pw reset page and then the datasheet page gives a valid PDF
- [ ] Test that after creating a user the same thing holds.

PR #167, opened 2018-03-14
Added cache group support. This closes #113

### LINKED ISSUE #113: Manage cache opt-in
We should at least display if a user has opted-in for the cache, even better would be the possibility to toggle that as in the Hausübersicht.

Paktosan:


Paktosan:
Awaiting your approval, @lukasjuhrich

lukasjuhrich:
I'm aware of that, and the Github mail remains being marked unread. Please note that this is not the only issue in my queue and it does not make any difference when I'll merge that, so it does not have a higher priority than other things.

tests/fixtures/config.py
```diff
@@ -106,3 +110,8 @@ class violation_network_access_deny:
         property_group = PropertyGroupData.violation
         name = "network_access"
         granted = False
+
+    class cache:
+        property_group = PropertyGroupData.cache
+        name = "cache_access"
+        granted = False
```
lukasjuhrich: This should be granted. Note that `property` does not say what properties exist, but which property groups grant or deny which properties. Properties come to life by being granted or denied.
I'm leaving this because it has no functional relevance in the tests (and these are the legacy fixtures).

PR #172, opened 2018-03-29
Hades

### LINKED ISSUE #101: Create PSQL-Views for Hades
Pycroft has to provide the Radius-Tables for Hades.

This includes (example taken from abe):

- `hades_dhcphost` (view)
- `hades_nas` (table)
- `hades_nasipaddress` (table)
- `hades_radcheck` (view)
- `hades_radgroupcheck` (table)
- `hades_radgroupreply` (table)
- `hades_radreply` (table)
- `hades_radusergroup` (view)

The creation of these tables should be Implemented using `SQLAlchemy` so `metadata.create_all` provides these tables as well.
This way, we can directly use these to run integration tests.

lukasjuhrich:
Fixes #101 

Left to do (following @sebschrader):

- [x] Remove Reply-Message reply items
- [x] Use case-sensitive column names for Hades views
- [x] Direct mapping from property name to group names
- [x] Use better names for the blocking properties (traffic_limit_exceeded, payment_in_default, no_network_access)
- [x] Add radgroupreply items for blocking properties
- [x] Set Fall-Through to yes for regular groups
- [x] Add (&& fill) the `radius_property`, `radgroupreply_base`
      and `nas` tables (the latter without secret)
- [x] Add the blocking groups in the importer
- [x] Add integration tests for each table, as started
- [x] Add view `dhcphost`: `Mac | IpAddress` && test
- [x] Add view `alternative_dns`: `IpAddress` && test

### DIFF
--- a/helpers/interactive.py
+++ b/helpers/interactive.py
@@ -27,8 +27,10 @@ from pycroft import config
 
 connection_string = os.environ['PYCROFT_DB_URI']
 
+engine = create_engine(connection_string, echo=True)
+
+
 def setup():
-    engine = create_engine(connection_string, echo=True)
     # TODO: don't set a scoped session, just a regular one
     set_scoped_session(scoped_session(sessionmaker(bind=engine),
                                       scopefunc=lambda: _request_ctx_stack.top))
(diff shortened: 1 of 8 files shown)


lukasjuhrich:
In the current `radgroupreply`, I noticed `unknown` and `valid` entries setting 2hades-unauth as well. I assume I should add those static entries to the view?

PR #183, opened 2018-05-23
Membership Fee

### LINKED ISSUE #179: Implement membership fee
The currently used membership contribution model is not implemented yet. However, the "semester fee" could be transformed to a general fee, because it already includes nearly all features we require.

This is probably required to implement #166.

ibot3:
- [ ] Rework "Semester Fee" to "Membership Fee" (Web, Lib, Model, Tests)
- [ ] Implement payment in default handler
- [ ] Add web pages for booking fees and handling payments in default

Closed #179


ibot3:
- [ ]  Prevent errors when no fees were created yet (post_fees, handle_payments_in_default)

lukasjuhrich:
- [ ] Add actual booking routine w/ the proposed checks

I met with @ibot3 and we agreed on the following workflow:
- [ ] add an exclusion constraint to `transaction` so every user can have at maximum one transaction per fee
- [ ] don't book all fees at once, but individually
- [ ] before booking the fee (adding the transactions), select all users having had property `membership_fee` at the effective date of the fee (which are not in the grace period) and display them in the list
- [ ] Implement the query that inserts a transaction (to all the people with the property & not in the grace period, as above) using `ON CONFLICT IGNORE`
- [ ] If it's possible to find out how many things have been ignored, simulate the query and display the value

lukasjuhrich:
OK, so filtering for the grace fee is a bit nontrivial, since one has to select all memberships (in groups that grant `membership_fee`, ignoring that it might be denied by some other membership) and intersect the union of their intervals with `[begin of fee, begin of fee + grace_period]` and see if it is non-empty or not, and use that as a `WHERE` condition on the user.

ibot3:
@lukasjuhrich Checking the membership status at the end of the fee period is not enough. We also want to book a fee if the user was a member from 01.x.y until 25.x.y. That means that we somehow need the amount of days a user was a member in the period. Do you have an idea how this could be done with psql?

lukasjuhrich:
@ibot3 I did not propose checking for membership at the end of the month, but at the start of the month – hence `[begin of fee, begin of fee + grace_period]`. This would be correct if I didn't miss anything.

tests/lib/test_finance.py
```diff
@@ -380,82 +295,10 @@ def test_grace(self):
...
-    def test_booked_fee_paid_too_late_with_late_fee_already_booked(self):
-        transaction = self.book_a_fee()
-        late_fee = self.late_fee_for(transaction)
-        simple_transaction(late_fee[0], self.fee_account,
-                           self.user.account, late_fee[1], self.user,
-                           late_fee[2])
-        self.pay_fee(self.payment_deadline + timedelta(days=1))
-        session.session.commit()
-        self.assertEqual(self.fee.compute(self.user),
-                         [self.late_fee_for(transaction)])
```
lukasjuhrich: Aahhhh, beautiful! Less pycroft is more pycroft! :heart:

Issue #162, opened 2018-02-19
Switch to timestamp with timezone


sebschrader:
PostgreSQL's two timstamp data types can be quite confusing. It took me a while to fully grok their semantics and find out, which type should be used. There are also a lot of misleading blog posts out there of people advocating to never use `timestamptz` and others advocating to never use `timstamp`.

The main source of confusion is, that `timestamp with timezone` *doesn't actually store any timezone information* at all. **Timestamp with timezone are always UTC!**. For `timestamptz` PostgreSQL uses the `timezone` of the database connection to convert from and to UTC. For `timestamp` the values are untouched.

So the use case for these two data types is as follows:

- `timestamptz` for events at a point in time
- `timestamp` for events at a moment in a not yet specified timezone. E.g. do something every Friday 20:00, but the exact time depends on the user.

So Pycroft basically does it all wrong, and uses `timestamp` everywhere although it should use `timestamptz`. Using `timestamp` required kludges, such as `utcnow()` (mentioned in #101). They are not needed with `timestamptz`.

lukasjuhrich:
Doesn't that just constitute
- [ ] Switching to `timestamp with timezone` everywhere in the model
- [ ] Switching from `functions.utcnow()` to `datetime.utcnow()` in the library
- [ ] Replacing usages of `functions.utcnow()` in expression form by the appropriate postgres function returning a `timestamptz` and not a `timestamp` object,

or am I missing something?

sebschrader:
Not completely. `psycopg2` will return timezone-aware instead of naïve `datetime.datetime` objects. By default, it uses its own [`timezone`](http://initd.org/psycopg/docs/tz.html#psycopg2.tz.FixedOffsetTimezone) class however for the `tz_info` property of the `datetime.datetime` objects. We probably don't want that and want always UTC in the form of `datetime.timezone.utc`, but AFAIK that wouldn't be strictly necessary.

Remember, that `timestamp with timezone` doesn't actually store a timezone, so what even is the timezone set by `psycopg2`? It is the timezone of the database connection.

[...]

PR #184, opened 2018-05-23
Retrieve bank account transactions via FinTS

### LINKED ISSUE #182: Retrieve bank account transactions via FinTS
Pycroft should be able to retrieve new bank account transaction directly from the bank. This can be done using the Finance Transaction System (FinTS). There already exists a FinTS library for python that we could use: https://github.com/raphaelm/python-fints

marcelb98:
Bank account transactions can now be imported via FinTS. The credentials for logging into the bank account are not saved anywhere (except the FinTS-URI).
Before importing all potentially new transactions, they're checked if they are new transactions. Potentially new are those transactions, which took place on the day of the last import or later.

With this PR also a bug, which occurred while listing all existing bank accounts when one of those hasn't had any transactions, was fixed.

Closes #182

### DIFF
--- a/web/blueprints/finance/__init__.py
+++ b/web/blueprints/finance/__init__.py
@@ -134,23 +134,23 @@ def bank_accounts_import():
     old_transactions = [] # transactions which are already imported
     if form.validate_on_submit():
         # login with fints
-        bankaccount = BankAccount.q.get(form.account.data)
+        bank_account = BankAccount.q.get(form.account.data)
         try:
             fints = FinTS3PinTanClient(
-                bankaccount.routing_number,
+                bank_account.routing_number,
                 form.user.data,
                 form.pin.data,
-                bankaccount.fints_endpoint
+                bank_account.fints_endpoint
             )
 
             acc = next((a for a in fints.get_sepa_accounts()
-                        if a.iban == bankaccount.iban), None)
+                        if a.iban == bank_account.iban), None)
             if acc is None:
                 raise KeyError('BankAccount with IBAN {} not found.'.format(
-                    bankaccount.iban)
+                    bank_account.iban)
                 )
-            if bankaccount.last_updated_at is not None:
-                start_date = bankaccount.last_updated_at.date()
+            if bank_account.last_updated_at is not None:
+                start_date = bank_account.last_updated_at.date()
             else:
                 start_date = date(2018,1,1)
             statement = fints.get_statement(acc, start_date, date.today())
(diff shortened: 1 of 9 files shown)


web/blueprints/finance/__init__.py
```diff
@@ -123,18 +129,92 @@ def actions(activity_id):
...
+                transaction.data['applicant_name'] is not None else ''
+            new_activity = BankAccountActivity(
+                bank_account_id=bankaccount.id,
+                amount=int(transaction.data['amount'].amount*100),
+                reference=transaction.data['purpose'],
+                original_reference=transaction.data['purpose'],
+                other_account_number=iban,
+                other_routing_number=bic,
+                other_name=other_name,
+                imported_at=datetime.now(),
```
Georgeto: I think it would be better to use the same timestamp for the whole batch of imported transactions.

Apart from that I am not sure if `datetime.now()` is the right way to get the current time. With the changes proposed in #162 it certainly wouldn't, but as the changes are still to be implemented, and me being not aware of the current timezone situation, it would be great, if a more seasoned Pycroft developer (e.g. @sebschrader) could comment on this.

web/blueprints/finance/__init__.py
```diff
@@ -123,18 +129,92 @@ def actions(activity_id):
...
+                    acc = account
+                    break
+            if acc is None:
+                raise KeyError('BankAccount with IBAN {} not found.'.format(
+                    bankaccount.iban)
+                )
+            if bankaccount.last_updated_at is not None:
+                start_date = bankaccount.last_updated_at.date()
+            else:
+                start_date = date(2018,1,1)
```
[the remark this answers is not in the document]
marcelb98: For fetching the statements of a bank account FinTS requires a starting date. The official FinTS-specification states that the date can be any legal ISO 8601 date (which includes 0001-01-01). But trying 0001-01-01 as the starting date with an endpoint of my local Sparkasse returns an error. The first working date is 0010-01-01.
Since I don't know how other banks are handling this date, I decided to use 2018-01-01 as the starting date if no import was done in the past. I think this is ok because all previous transactions should be imported through the importer-scripts.

web/blueprints/finance/__init__.py
marcelb98: (This comment is yet not outdated.)

marcelb98: I agree that it would be better to use the same timestamp for the whole batch. But before changing this in code I'll wait for comments on the way of getting the current time.

Issue #165, opened 2018-03-12
Create routine to block users with exceeded traffic


lukasjuhrich:
This includes:

- [ ] Adding a “Blocked (Traffic)” PropertyGroup to the config (which grants `blocking_reason_traffic`
- [ ] Looking if you want to implement #129 
- [ ] Making this routine available as a `console_script` in the `setup.py` so regular execution using cron or systemd timers becomes easy

sebschrader:
In a similar vein to #166, the property/group should be called `traffic_limit_exceeded` and not blocked by etc. Pycroft doesn't block anyone, it just detects various conditions and sets group memberships, if those are met.

It's upon us (I.e. policy) to decide to revoke network access permissions, thus blocking, in this case, but we might not do it.

lukasjuhrich:
The filtering of all the users can be done by `User.q.join(User._current_traffic_balance).filter(CurrentTrafficBalance.amount<0).all()`.
Note that this does not look at whether the user is actually blockable for traffic or not.

PR #187, opened 2018-06-11
Traffic limit exceeded

ibot3:
Implements routine to add the `traffic_limit_exceeded` group to members who exceeded their traffic limit.


setup.py
```diff
@@ -25,6 +25,7 @@
             'pycroft_legacy_cache = legacy.cache_legacy:main',
             'pycroft_legacy_gerok_import = legacy_gerok.__main__',
             'pycroft_ldap_sync = ldap_sync.__main__',
+            'pycroft_handle_exceeded_traffic_limits = traffic.handle_exceeded_traffic_limits',
```
ibot3: @lukasjuhrich 
Does this automaticly commit the transaction which was started by the called function?

https://github.com/agdsn/pycroft/pull/187/files#diff-f9284d5e7e7a9b14181de8f6d0c99416R160

pycroft/lib/traffic.py
```diff
@@ -151,3 +155,30 @@ def reset_credit(user, processor, target_amount=1*1024**3):
     log_user_event(deferred_gettext("Traffic has ben compensated to 1GiB").to_json(),
                    author=processor,
                    user=user)
+
+
+@with_transaction
+def handle_exceeded_traffic_limits():
```
lukasjuhrich: As a rule of thumb, `handle` is not a good name for describing an action. It could mean anything, and furthermore, you don't even describe what object is subject to handling (the user, which gets or a property).

I propose `sync_traffic_blockages` and adding a docstring (better two for blocking/unblocking) describing the technicalities (a membership in $foo is added if and only if…).

pycroft/lib/traffic.py
```diff
@@ -151,3 +155,30 @@ def reset_credit(user, processor, target_amount=1*1024**3):
...
+    # End memberships
+    users = User.q.join(User._current_traffic_balance) \
+        .filter(or_(CurrentTrafficBalance.amount >= 0,
+                    User.has_property('traffic_limit_disabled'))) \
+        .filter(User.has_property('traffic_limit_exceeded')) \
+        .all()
+
+    for user in users:
+        remove_member_of(user, config.traffic_limit_exceeded_group,
+                         processor, closed(session.utcnow(), None))
```
lukasjuhrich: Same concerns as above.

setup.py
lukasjuhrich: I'm actually not sure the decorator works, so no Idea. Why'd you ask, did you plan on committing this feature untested? ;)

tests/lib/test_traffic.py
```diff
@@ -55,6 +65,40 @@ def test_0030_expr_comparator(self):
         # test comparator expression
         correct_values = [(u.id, b > 0 if b is not None else None)
                           for u, b in self.correct_balance.items()]
-        res = session.session.query(User.id, traffic_balance_expr()>0).all()
+        res = session.session.query(User.id, traffic_balance_expr() > 0).all()
 
         self.assertEqual(set(correct_values), set(res))
+
+
+class Test_020_TrafficLimitExceeded(FixtureDataTestBase):
```
lukasjuhrich: :+1: for unprompted testing.
Remember to validate your test by _removing parts of the logic_ (e.g., removing a `has_property` filter) and confirming that the test actually fails! Else it's just a meaningless green light.

pycroft/lib/traffic.py
ibot3: I wanted to avoid the word "block" because of what shreyder [wrote at issue 165](https://github.com/agdsn/pycroft/issues/165#issuecomment-372234114).

scripts/sync_exceeded_traffic_limits.py
```diff
@@ -0,0 +1,42 @@
...
+        connection_string = os.environ['PYCROFT_DB_URI']
+    except KeyError:
+        raise RuntimeError("Environment variable PYCROFT_DB_URI must be "
+                           "set to an SQLAlchemy connection string.")
+
+    engine = create_engine(connection_string)
+    connection = engine.connect()
+    state = AlembicHelper(connection)
+    strategy = SchemaStrategist(state).determine_schema_strategy()
+    strategy()
```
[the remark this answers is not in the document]
ibot3: What would be the correct way to do this?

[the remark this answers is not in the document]
ibot3: @lukasjuhrich ?

lukasjuhrich:
Fixes #165

PR #192, opened 2018-08-01
Switch to timestamp with timezone

Georgeto:
Task list taken from the corresponding issue #162:
- [x] Set the timezone connection parameter to `Etc/UTC`
- [x] Switch to `timestamp with timezone` everywhere in the model
- [x] Switch from `pycroft.model.functions.utcnow()` to `datetime.datetime.now(timezone.utc)` in the library
- [x] Replace usages of `pycroft.model.functions.utcnow()` in SQL expressions by `sqlalchemy.func.current_timestamp()`
- [ ] (Maybe optional, but recommended) Instruct `psycopg2` to return `datetime.timezone.utc`
- [ ] Verify the output of date and time in the web interface. Flask-Babel's [`format_datetime`](https://pythonhosted.org/Flask-Babel/#flask.ext.babel.format_datetime) should already do all the hard work. I think we're using it.

### DIFF
--- a/pycroft/model/user.py
+++ b/pycroft/model/user.py
@@ -463,12 +463,14 @@ class Membership(IntegerIdModel):
                 "begins_at must be before ends_at"
         return value
 
-    def disable(self):
-        now = session.utcnow()
-        if self.begins_at is not None and self.begins_at > now:
+    def disable(self, ends_at=None):
+        if ends_at is None:
+            ends_at = session.utcnow()
+
+        if self.begins_at is not None and self.begins_at > ends_at:
             self.ends_at = self.begins_at
         else:
-            self.ends_at = now
+            self.ends_at = ends_at
 
 
 class PropertyGroup(Group):
--- a/pycroft/model/user.py
+++ b/pycroft/model/user.py
@@ -15,9 +15,9 @@ import re
 
 from flask_login import UserMixin
 from sqlalchemy import (
-    Boolean, BigInteger, CheckConstraint, Column, DateTime, ForeignKey, Integer,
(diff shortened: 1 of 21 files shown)


Georgeto:
The failing membership tests assume that time returned by `session.utcnow()` changes between the call to `Membership.disable()` and `Membership.active()`. That was the case with the previous implementation of `session.utcnow()`, which used `STATEMENT_TIMESTAMP()` (Current date and time at the **start of current statement**) as its timestamp provider. However, the new implementation uses  `CURRENT_TIMESTAMP` (Current date and time at the **start of current transaction**) as its time provider. All tests run inside their own nested transactions, but `CURRENT_TIMESTAMP` seems to return the start timestamp of the outermost transaction. Consequently all calls to `CURRENT_TIMESTAMP` return the same value, which leads to the failing membership tests.

I am unsure what would be the best way to fix the tests. Manually set appropriate `begins_at`/`ends_at` timestamps, instead of just assigning  `session.utcnow()`? Let  `session.utcnow()` use `STATEMENT_TIMESTAMP()` instead of  `CURRENT_TIMESTAMP`?

pycroft/model/user.py
```diff
@@ -416,7 +417,7 @@ def active(self, when=None):
         :rtype: bool
         """
         if when is None:
-            now = object_session(self).query(functions.utcnow()).scalar()
```
sebschrader: There is a reason, why `object_session` was used to obtain the correct session. Please don't break this.

ibot3:
`[begins_at, ends_at]` for memberships is fine in my opinion.

PR #199, opened 2018-08-14
[LDAP] Sync whether login should be disabled

### LINKED ISSUE #198: Export blocked users to LDAP as well
Following the discussion in #126

lukasjuhrich:
Attempts to fix #198


lukasjuhrich:
Now actually fixes #198 :p

Issue #224, opened 2018-08-31
Revise subnet address reservation


Georgeto:
Allow reservation of addresses at the bottom and the top of a subnet, instead of only at the bottom.
We need this for example for the Zeunerstraße, as the upper part of the subnet is used for other purposes through the use of proxy arp.

- [x] Divide `Subnet.reserved_addresses` into `reserved_addresses_bottom` and `reserved_addresses_top`
- [x] Adapt `get_free_ip` to the new subnet model
- [x] Write a alembic migration for the model changes

PR #225, opened 2018-09-01
Revise address reservation

Georgeto:
see #224

### DIFF
--- a/web/blueprints/infrastructure/__init__.py
+++ b/web/blueprints/infrastructure/__init__.py
@@ -36,6 +36,32 @@ def subnets():
         subnet_table=SubnetTable(data_url=url_for(".subnets_json")))
 
 
+def format_address_range(base_address, amount):
+    if amount == 0:
+        raise ValueError
+    elif abs(amount) == 1:
+        return str(base_address)
+    elif amount > 1:
+        return '{} - {}'.format(str(base_address),
+                                str(base_address + amount - 1))
+    else:
+        return '{} - {}'.format(str(base_address + amount + 1),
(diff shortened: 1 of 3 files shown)


lukasjuhrich:
Does the alembic revision control work? This would be the first time we had t use it in production.

pycroft/lib/net.py
```diff
@@ -15,9 +16,14 @@ class MacExistsException(Exception):
...
     for subnet in subnets:
-        reserved = subnet.reserved_addresses or 0
+        reserved_bottom = subnet.reserved_addresses_bottom or 0
+        reserved_top = subnet.reserved_addresses_top or 0
         used_ips = frozenset(ip.address for ip in subnet.ips)
-        unreserved = islice(subnet.address.iterhosts(), reserved, None)
+        unreserved = islice(
+            subnet.address.iterhosts(), reserved_bottom,
+            # Stop argument must be None or an integer: 0 <= x <= sys.maxsize.
+            # IPv6 subnets can exceed this boundary on 32 bit python builds.
```
lukasjuhrich: Oh, good catch, would never have thought about that. :+1:

lukasjuhrich:
As CI is broken, I'm assuming you ran the relevant unittests locally?

Georgeto:
> Does the alembic revision control work? This would be its first application.
With the fix in 139e3dd, the migration applied successfully on my local database.

> As CI is broken, I'm assuming you ran the relevant unittests locally?
Yes, I did.

lukasjuhrich:
Splendid! Feel free to merge.

PR #246, opened 2018-09-17
Hosts and interfaces

### LINKED ISSUE #243: Allow editing of hosts/interfaces/ip
It should be possible to edit/add/remove hosts, interfaces and IPs of a user.
This is probably a requirement of #242.

ibot3:
Reworks the one "Hosts" table on the user_show page to two tables ("Hosts" & "Interfaces"). Hosts, Interfaces and IPs can be added/edited/deleted with these tables.

With the ability of manually adding hosts + interfaces, it is now possible to not supply the MAC address on user creation (room and birthdate will still be saved).

This also adds a generic form template which could be used of various endpoints in the future which only contain simple forms.

Closes #243


web/blueprints/user/tables.py
```diff
@@ -72,17 +72,59 @@ def generate_toolbar(self):
...
+            Column('edit_link', 'Editieren', formatter='table.btnFormatter'),
+            Column('delete_link', 'Löschen', formatter='table.btnFormatter')
+        ], **kw)
+
+        self.user_id = user_id
+
+    def generate_toolbar(self):
+        if self.user_id is None:
+            return
+        #if not current_user.has_property('user_hosts_change'):
```
Georgeto: Why is the property check disable?

web/blueprints/user/tables.py
```diff
@@ -72,17 +72,59 @@ def generate_toolbar(self):
...
+            Column('ips', 'IPs'),
+            Column('edit_link', 'Editieren', formatter='table.btnFormatter'),
+            Column('delete_link', 'Löschen', formatter='table.btnFormatter')
         ], **kw)
+        self.user_id = user_id
+
+    def generate_toolbar(self):
+        if self.user_id is None:
+            return
+        #if not current_user.has_property('user_hosts_change'):
```
Georgeto: And again, why is the property check disable?

pycroft/lib/net.py
```diff
@@ -24,11 +26,24 @@ def get_free_ip(subnets):
             # Stop argument must be None or an integer: 0 <= x <= sys.maxsize.
             # IPv6 subnets can exceed this boundary on 32 bit python builds.
             min(subnet.address.numhosts - reserved_top - 2, sys.maxsize))
-        unused = (ip for ip in unreserved if ip not in used_ips)
-        try:
-            return next(unused), subnet
-        except StopIteration:
-            continue
+        unused = unused + list(ip for ip in unreserved if ip not in used_ips)
```
Georgeto: Eagerly constructing a list of all unused IPs is not optimal, using a generator would be preferable:
```python
for ip in unreserved:
    if ip not in used_ips:
        yield ip
```
OR
```python
for unused_ip in (ip for ip in unreserved if ip not in used_ips):
    yield unused_ip
```

pycroft/lib/net.py
```diff
@@ -24,11 +26,24 @@ def get_free_ip(subnets):
...
+
+
+def get_free_ip(subnets):
+    unused = iter(get_unused_ips(subnets))
+
+    try:
+        ip = next(unused)
+
+        subnet = next(
+            iter([subnet for subnet in subnets if (ip in subnet.address)]),
```
Georgeto: Again, a generator comprehension would be preferable:
```
subnet = next(
            (subnet for subnet in subnets if (ip in subnet.address)),
            None)
```

Georgeto:
When creating/editing a host for a user, it is impossible to select the room the host is located in, but instead the room is automatically set to the user's room. While this might be appropriate in most cases, it should be possible to create hosts that are not located in the user's room.

PR #266, opened 2018-10-09
Fix spelling

JuKu:
Fix spelling

### DIFF
--- a/web/templates/user/_user_show_traffic.html
+++ b/web/templates/user/_user_show_traffic.html
@@ -1,4 +1,4 @@
-{#
+﻿{#
  Copyright (c) 2014 The Pycroft Authors. See the AUTHORS file.
  This file is part of the Pycroft project and licensed under the terms of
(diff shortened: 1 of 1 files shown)


web/templates/user/_user_show_traffic.html
```diff
@@ -1,4 +1,4 @@
-{#
```
Georgeto: Why is there a change here? Did you add/remove a byte-order mark?

ibot3: @JuKu 
There is still a change there for no reason.

JuKu:
@sebschrader @Georgeto Is there any progress here?
Or can we close / delete this PR?

lukasjuhrich:
@JuKu as has been mentioned [here](https://github.com/agdsn/pycroft/pull/266/files#r243905357), it still contains independent changes. please amend that, and force-push your `develop` branch.

PR #268, opened 2018-10-09
Fix RuntimeError on empty Radius-Logs

strebbin:
Log for Users in dormitories RBS and Gutz show HTTP Error 500 because the Switches in these dormitories are not able to do radius auth. If there Is an empty answer from Hades, it causes an HTTP 500.


Georgeto:
Thank you!

PR #271, opened 2018-10-11
Improve PyCharm integration

Georgeto:


sebschrader:
The `rest.xml` could remain. Why did you delete it?

Georgeto:
> The `rest.xml` could remain. Why did you delete it?

I have mistakenly assumed that the `rest.xml` is related to PyCharm's "Test RESTful web service" feature.

PR #272, opened 2018-10-11
Adapt travis configuration to reworked docker setup

Georgeto:
The failing tests are hades and ldap sync tests that also fail when executed locally.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -14,6 +14,7 @@ env:
     - DOCKER_CACHE_FILE=/home/travis/docker/cache.tar.gz
     - DOCKER_COMPOSE="docker-compose -f docker-compose.test.yml"
     - DOCKER_COMPOSE_VERSION=1.22.0
+    - UID=$TRAVIS_UID
 
 before_install:
   # TODO: Remove when travis has a adequate version preinstalled
--- a/.travis.yml
+++ b/.travis.yml
@@ -12,7 +12,7 @@ services:
 env:
   global:
     - DOCKER_CACHE_FILE=/home/travis/docker/cache.tar.gz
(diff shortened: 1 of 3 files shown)


.travis.yml
```diff
@@ -12,9 +12,22 @@ services:
...
+  # upgrade to latest docker version to avoid "Forbidden path outside the build context: ../Dockerfile" error
+  - curl -fsSL https://download.docker.com/linux/ubuntu/gpg | sudo apt-key add -
+  - sudo add-apt-repository "deb [arch=amd64] https://download.docker.com/linux/ubuntu $(lsb_release -cs) stable"
+  - sudo apt-get update
+  - sudo apt-get -y install docker-ce
+  # upgrade to latest docker-compose
+  - sudo rm /usr/local/bin/docker-compose
+  - curl -L https://github.com/docker/compose/releases/download/${DOCKER_COMPOSE_VERSION}/docker-compose-`uname -s`-`uname -m` > docker-compose
+  - chmod +x docker-compose
+  - sudo mv docker-compose /usr/local/bin
```
lukasjuhrich: This might be better placed in a `install_latest_docker.sh` script to not cludder the `.travis.yml`.

Georgeto: > lukasjuhrich marked this conversation as resolved.

?

lukasjuhrich: I retroactively thought `comment` would have been better than `request changes`, but you can't change that. So I just marked it as resolved so it wouldn't be a blocker. Which also didn't work.

Georgeto: It was a reasonable suggestion, and the fix was easy enough, so I guess requesting changes was fine.

PR #276, opened 2018-10-17
2 users1 room v2

Georgeto:
I added an page that lists all multi occupancies, or to be more precise, all rooms containing multiple users each of which has a host in the room.

@ibot3: You requested an "auto move out" feature based on the move in date. Could you elaborate on that?


ibot3:
Closes #69

PR #280, opened 2018-10-24
WIP: Keyboard shortcuts

Paktosan:


toothstone:
Not sure if you are already looking for feedback, but there seems to be an awful amount of unnecessary diff due to white space changes in the first commit (web/templates/layout.html), I would suggest reverting that before you continue.

MarauderXtreme:
Try to prevent your IDE from auto formating the source code file you are currently editing. As @toothstone said this shouldn't be here.

web/resources/js/spotlight.js
```diff
@@ -0,0 +1,389 @@
+// TODO:
```
lukasjuhrich: Please remove these comments and add appropriate issues.

web/resources/js/spotlight-modules.js
```diff
@@ -0,0 +1,111 @@
...
+        var populatedResults = ''
+        // console.log(forecast)
+        var forecast = forecast.concat(forecast)
+        $(forecast).each(function () {
+            var image = {}
+            image.alt = this.text
+
+            if (!["Snow Showers", "Rain And Snow", "Rain", "Partly Cloudy", "Mostly Cloudy", "Cloudy", "Showers", "Scattered Showers", "Scattered Thundershowers", "Mostly Sunny", "Thunderstorms", "Sunny", "Clear", "Snow", "Scattered Thunderstorms", "Breezy", "Mostly Clear", "Scattered Snow Showers", "Isolated Thunderstorms", "Windy", "Wintry Mix", "Blowing Snow"].includes(this.text)) {
+                alert(this.text);
+            }
```
lukasjuhrich: Please clean up examples. ;)

web/resources/js/keyboard-shortcuts.js
```diff
@@ -0,0 +1,18 @@
...
+hotkeys('option+f', function (event) {
+    window.location = "/user/search";
+});
+hotkeys('shift+option+f', function (event) {
+    window.open("/user/search");
+});
+hotkeys('option+h', function (event) {
+    window.open("https://www.youtube.com/embed/Ita6WyQ3SFQ?autoplay=1&controls=0");
+});
+
```
lukasjuhrich: I know this is personal taste, but I would prefer using a no-shiftkey-scheme like github uses: `option` (or `alt`) can clash with e.g. window managers, while just typing `p s` („pycroft search“) or whatever will always work (except on input fields, but that's restricted even with prefixes appearently).

Lodifice: I accept your critique regarding window manager shortcuts, but the ones I know mostly use the mod key (Window/Tux symbol, depending on the keyboard). Additionally, simple key commands will likely conflict with browser plugins that allow sane surfing (e.g. Vimium and Tridactyl).

The best option is of course to finish the API so that an ncurses client may be developed :D

lukasjuhrich: @Lodifice I believe KDE has some things bound to `Meta-*`, and apart from that it is the default option for triggering menu hotkeys. If I press `M-h` on this very page on an uncustomized firefox-developer installation, it opens the “help” menu. Besides that, I know more websites that embrace a single-key setup than using a prefix – I make heavy use of this in e.g. tweetdeck, github and gitlab. If you use vimperator, wouldn't you find a scheme like this more intuitive? Wouldn't it be possible to avoid clashes by using a prefix like `p`, `x` or similar?

> The best option is of course to finish the API so that an ncurses client may be developed :D
Indeed.

mmarx: Don't confuse `M-h` and `<Alt> h` (the latter does indeed open the “help” menu on my Firefox 63); `hotkeys-js` does not seem to support proper Meta keys. Furthermore, don't break keybindings in my browser: Github, among others, rebinds `/`, and Tridactyl contains special-case code to unbreak searching for text in pages.

sebschrader:
As this feature is a bit controversial, please add the ability to disable it. The user's setting should be persisted somehow (probably through a long-lived cookie) and should always be disabled when not logged in.

lukasjuhrich:
@Paktosan any updates on this?

Paktosan:
This is pretty much dead and I currently have no intentions of picking it up again.

PR #283, opened 2018-10-26
Display number of free IPs in subnet overview

Georgeto:
image


mueslo:
The formatting is needlessly confusing. Why not just display "16 (of 242)" or just show how many IPs are in a subnet next to the subnet mask.

Georgeto:
> The formatting is needlessly confusing. Why not just display "16 (of 242)" or just show how many IPs are in a subnet next to the subnet mask?

You are right, I have changed it to "16 (of 242)", thank you.

PR #293, opened 2018-12-20
Error handling for finance importer

marcelb98:
This PR adds:

- switch to python-fints version 2.0.0
- catch errors which occur during finance import and save them while importing valid statements
- ability to review and fix mt940 of errored statements

**In detail:**
If a part of the MT940 received from our bank is not valid, this part will be skipped and saved with the exception. After importing all other statements the treasurer has the ability to review the MT940, fix it and execute the importer again.
The UI for fixing the MT940 could be found at "Finanzen > Fehlerhafter Bankimport".

### DIFF
--- a/pycroft/model/finance.py
+++ b/pycroft/model/finance.py
@@ -325,6 +325,25 @@ class BankAccountActivity(IntegerIdModel):
     )
 
 
+class MT940Error(IntegerIdModel):
+    mt940 = Column(Text(), nullable=False)
+    exception = Column(Text(), nullable=False)
+    author = relationship("User")
+    author_id = Column(Integer, ForeignKey("user.id"), nullable=False)
+    imported_at = Column(DateTimeTz, nullable=False,
+                       server_default=func.current_timestamp(),
+                       onupdate=func.current_timestamp())
+    bank_account = relationship(BankAccount, backref=backref("mt940_errors"))
+    bank_account_id = Column(Integer, ForeignKey(BankAccount.id),
+                             nullable=False)
+
+    def __init__(self, mt940, exception, author, bank_account):
+        self.mt940 = mt940
+        self.exception = exception
+        self.author = author
+        self.bank_account = bank_account
+
+
 manager.add_constraint(
     BankAccount.__table__,
     CheckConstraint(
--- a/web/blueprints/finance/__init__.py
+++ b/web/blueprints/finance/__init__.py
@@ -33,11 +33,12 @@ from pycroft.lib.finance import get_typed_splits, \
     post_transactions_for_membership_fee, build_transactions_query, \
     match_activities, take_actions_for_payment_in_default_users
 from pycroft.model.finance import (
-    BankAccount, BankAccountActivity, Split, MembershipFee)
+    BankAccount, BankAccountActivity, Split, MembershipFee, MT940Error)
 from pycroft.model.session import session
 from pycroft.model.user import User
 from pycroft.model.finance import Account, Transaction
 from web.blueprints.access import BlueprintAccess
+from web.blueprints.helpers.fints import FinTS3Client
 from web.blueprints.helpers.table import date_format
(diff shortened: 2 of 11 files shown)


pycroft/model/finance.py
```diff
@@ -325,6 +325,25 @@ class BankAccountActivity(IntegerIdModel):
...
+    author = relationship("User")
+    author_id = Column(Integer, ForeignKey("user.id"), nullable=False)
+    imported_at = Column(DateTimeTz, nullable=False,
+                       server_default=func.current_timestamp(),
+                       onupdate=func.current_timestamp())
+    bank_account = relationship(BankAccount, backref=backref("mt940_errors"))
+    bank_account_id = Column(Integer, ForeignKey(BankAccount.id),
+                             nullable=False)
+
+    def __init__(self, mt940, exception, author, bank_account):
```
Georgeto: I don't think it is necessary to explicitly define the constructor.

web/blueprints/finance/forms.py
```diff
@@ -7,6 +7,7 @@
 from wtforms import Form as WTForm, ValidationError
 from wtforms.validators import DataRequired, NumberRange, Optional, \
     InputRequired
+from wtforms.widgets import TextArea
```
Georgeto: Why not use our decorated TextAreaField?
https://github.com/agdsn/pycroft/blob/ac8be4c9c49ac6ae5a005a08b152625e65921b39/web/form/fields/core.py#L192-L198

pycroft/model/finance.py
marcelb98: If the constructor is not defined explicitly, every attribute has to be set individually when creating a new entry.

web/blueprints/finance/__init__.py
```diff
@@ -172,9 +188,12 @@ def bank_accounts_import():
                     bank_account.iban)
                 )
             start_date = form.start_date.data
-            statement = fints.get_statement(acc, start_date, date.today())
+            statement, with_error = fints.get_filtered_transactions(acc, start_date, date.today())
             flash(
                 "Transaktionen vom {} bis {}.".format(start_date, date.today()))
+            if len(with_error) > 0:
+                flash("{} Statements enthielten fehlerhafte Daten und müssen vor dem Import manuell korrigiert werden.".format(len(with_error)), 'error')
```
lukasjuhrich: Pretty sure that's long enough to violate the style guidelines for line length (90 soft limit, 100 hard limit) as it even wraps in the github diff view.

web/blueprints/finance/__init__.py
```diff
@@ -137,6 +138,21 @@ def actions(activity_id):
...
+        {
+            'name': error.bank_account.name,
+            'fix': {
+                'href': url_for('.fix_import_error',
+                                error_id=error.id),
+                'title': 'korrigieren',
+                'btn_class': 'btn-primary'
+            },
+            'imported_at': '{}'.format(
+                map_or_default(error.imported_at, datetime.date, 'nie'))
```
lukasjuhrich: FYI, with the upcoming  [Py3.8 Assignment expressions](https://www.python.org/dev/peps/pep-0572/), the helper function could be inlined to
```python
'imported_at': str(datetime.date(i)) if (i := error.imported_at) is not None else 'nie'
```

web/blueprints/helpers/fints.py
```diff
@@ -0,0 +1,61 @@
...
+from fints.segments.statement import HKKAZ5, HKKAZ6, HKKAZ7
+from fints.utils import mt940_to_array
+
+logger = logging.getLogger(__name__)
+
+class FinTS3Client(FinTS3PinTanClient):
+
+    def get_filtered_transactions(self, account: SEPAAccount,
+                         start_date: datetime.date = None,
+                         end_date: datetime.date = None):
```
lukasjuhrich: Yay type hints! :smile: :tada:

Issue #126, opened 2017-12-04
Export properties as LDAP groups


lukasjuhrich:
The LDAP syncer should be able to sync a configurable subset of the granted (or denied) properties to `nestedGroupOfNames` so services can test for membership.

sebschrader:
I have never heard of `nestedGroupOfNames` and Google confirms that are probably is no such thing. You can always nest groups in LDAP by setting the member attribute to the DN of another group instead of a user.

Furthermore we should not use `groupOfNames` and use `groupOfMembers` from RFC2307bis instead. `groupOfNames` has some nasty disadvantages:

1. It's a structural rather than an auxiliary class, thus you can't combine it with other classes such as `posixGroup` that are structural too.

2. The member attribute is required, i.e. you must have at least one member per group.

lukasjuhrich:
@sebschrader Could you please address this question? It is important for the planning to know whether this feature is a “Bare Minimum” or can be postponed.

sebschrader:
>@sebschrader Could you please address this question? It is important
>for the planning to know whether this feature is a “Bare Minimum” or
>can be postponed.

The example you gave, can be addressed. The `aktiv` group can simply be converted into a *local UNIX group* on atlantis. Group membership in this group is already managed manually. Doing it in `/etc/group` or LDAP doesn't make much of a difference, if any.

There are other areas, where I don't have a good solution. Without manual managed blacklists on the respective system, we can only exclude users from or all or none of our services with the current mechanism by *denying the LDAP export property*. If the system distinguishes between “account not found” and other authentication errors, the user would temporarily get an “account not found” error.

In particular that means, we can't easily deny him login rights to `userlogin` while retaining his right to receive mails. Receiving mails is the most problematic case, I could think of. You don't want to temporarily reject someone's mails with a permanent error, because mails would immediately bounce. Not that we're making use of it very often, but you also couldn't use the user's mail address to contact him anymore.

If there is a way to address the mail issue or we can live with the fact, that we can't revoke someone's access to other services, that make use of LDAP, while retaining his mail service, we are fine otherwise it's a blocker.

mmarx:
How about using `pam_check_host_attr` to restrict login to certain hosts?

sebschrader:
> How about using `pam_check_host_attr` to restrict login to certain hosts?

[...]

PR #297, opened 2019-01-02
 Add support for groups and properties to LDAP sync 

Georgeto:


ibot3:
Closes #126

ldap_sync/exporter.py
```diff
@@ -5,28 +5,45 @@
...
+
+
+class GroupProxyType(NamedTuple):
+    Group: Group
+    members: List[str]
+
+
+class PropertyProxyType(NamedTuple):
+    name: str
+    members: List[str]
```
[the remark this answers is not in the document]
lukasjuhrich: Perhaps naming these types `PropertyResult` etc. would be sufficient, as imho being a result of some sort is closely tied to coming from a sql query in pycroft, already clarifying where it's being used.

ldap_sync/exporter.py
```diff
@@ -40,35 +57,65 @@ class LdapExporter(object):
...
+                    for x in db_users)
+
+        # Remove members that are not exported from groups/properties
+        exported_users = { u.User.login for u in db_users }
+
+        if db_groups:
+            desired.append(
+                GroupRecord.from_db_group(x.Group.name,
+                                          (m for m in x.members if m in exported_users),
+                                          group_base_dn, user_base_dn) for x in db_groups)
```
lukasjuhrich: Please indent the `for x in db_groups` part of the genexp on a same level as `GroupRecord.[…]`, because at the moment it may seem to be a genexp inside the `from_db_group()` call.

ldap_sync/exporter.py
```diff
@@ -40,35 +57,65 @@ class LdapExporter(object):
...
+        if db_properties:
+            desired.append(
+                GroupRecord.from_db_group(x.name,
+                                          (m for m in x.members if m in exported_users),
+                                          property_base_dn, user_base_dn) for x in db_properties)
```
lukasjuhrich: Indentation, same as above

ldap_sync/exporter.py
```diff
@@ -40,35 +57,65 @@ class LdapExporter(object):
...
+
```
lukasjuhrich: Are there tests ensuring that members are added and deleted properly when the group already exists?

ldap_sync/exporter.py
```diff
@@ -147,6 +184,57 @@ def fetch_users_to_sync(session, required_property=None):
...
+        # Grab all users with the required property
+        Group.q
+        # uids of the members of the group
+        .add_column(select([func.array_agg(User.login)])
+                .select_from(join(Group, Membership).join(User))
+                .where(Membership.active())
+                .group_by(Group.id)
+                .as_scalar().label('members'))
+        .all()
+    )
```
lukasjuhrich: This has probably been discussed and I'm out of the loop, but what is the rationale behind circumventing the permission system of pycroft and syncing groups instead of properties?
For instance, it does not seem possible to me to block a user from certain services without terminating the group membership. (CC @sebschrader )

ldap_sync/record.py
```diff
@@ -1,12 +1,19 @@
...
 from .action import AddAction, DeleteAction, IdleAction, ModifyAction
 
 
 def dn_from_username(username, base):
-    return "uid={},{}".format(username, base)
+    return safe_dn(["uid={}".format(username), base])
+
+
+def dn_from_cn(name, base):
+    return safe_dn(["cn={}".format(name), base])
```
lukasjuhrich: Good call with `safe_dn`

ldap_sync/exporter.py
```diff
@@ -5,28 +5,45 @@
...
 logger = logging.getLogger('ldap_sync')
 
 
+class UserProxyType(NamedTuple):
+    User: User
+    should_be_blocked: bool
+
+
+class GroupProxyType(NamedTuple):
+    Group: Group
```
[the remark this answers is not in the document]
Georgeto: Not my choice, the naming is dictated by SQLAlchemy.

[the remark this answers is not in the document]
lukasjuhrich: Ah, alright.

ldap_sync/exporter.py
Georgeto: Actually, both groups and properties are synced. But now that you mention it, I haven't thought whether we really need both, or if only syncing the properties would be enough. After looking at @sebschrader's config for OpenLDAP's SQL backend, I assumed that it was consensus that both groups and properties should be synced.

Georgeto:
@lukasjuhrich Are you satisfied with the rectifications that I have made?

PR #301, opened 2019-01-28
Remove traffic credit and groups

ibot3:
This PR removes all things that have to do with traffic credit/balance or traffic groups.
I removed lots of stuff which isn't required anymore for the productive usage. This makes the whole traffic logic much simpler.

I also introduced a list with the users that have the highest traffic usage.

This depends on https://github.com/agdsn/sipa/pull/396.


pycroft/lib/traffic.py
```diff
@@ -6,188 +6,14 @@
...
-                  .filter(or_(CurrentTrafficBalance.amount >= 0,
-                              CurrentProperty.property_name == 'traffic_limit_disabled')) \
-                  .all()
-
-    for user in users:
-        remove_member_of(user, config.traffic_limit_exceeded_group,
-                         processor, closed(session.utcnow(), None))
-        print("Traffic-Limit no longer exceeded: " + user.name)
\ No newline at end of file
+def get_users_with_highest_traffic(limit):
```
Georgeto: Did you intend to add this function?

ibot3: Yes, see the referenced issue above (#302).

Georgeto: But how is this related to this pull request?

ibot3: Not really. It's a function that we will need when removing the credit system, but it is not directly associated with the PR.

Should I remove it?

Georgeto: Making it part of the pull request for #302 would be better in my opinion.

pycroft/model/traffic.py
```diff
@@ -2,11 +2,13 @@
 # Copyright (c) 2015 The Pycroft Authors. See the AUTHORS file.
 # This file is part of the Pycroft project and licensed under the terms of
 # the Apache License, Version 2.0. See the LICENSE file for details.
+from datetime import timedelta
```
Georgeto: Seems to me like this import is unused.

web/api/v0/__init__.py
```diff
@@ -77,19 +76,14 @@ def get_interface_or_404(interface_id):
...
-        traffic_maxmium = None
-
     props = {prop.property_name for prop in user.current_properties}
     user_status = status(user)
 
     interval = timedelta(days=7)
     step = timedelta(days=1)
     traffic_history = func_traffic_history(user.id,
                                           session.utcnow() - interval + step,
-                                          interval, step)
```
Georgeto: Looks broken to me, as the new traffic_history function expects `(start_date, end_date)` instead of `(start_date, interval, step)`.

Georgeto: Whoever created the API should have written tests for it!

ibot3: Fixed.

pycroft/lib/traffic.py
ibot3: I just added the functionality here.

Georgeto:
A new version of SIPA, including the changes from agdsn/sipa#396, has been deployed. So I guess this PR can bee merged?

PR #306, opened 2019-02-14
Move name and birth date out of link so they are easier to copy

Paktosan:
In Chromium it is hard to mark the name and birth date because it will try to move the link itself. This makes that easier. Also there is no reason for the link to contain the complete text.

### DIFF
--- a/web/templates/user/_user_show_basedata.html
+++ b/web/templates/user/_user_show_basedata.html
@@ -18,7 +18,7 @@
 
             {% call info_row("Name") %}
                 {% if current_user is privileged_for('user_change') %}
-                <a href="{{ url_for(".edit_user", user_id=user.id) }}">{{ user.name }} <span class="glyphicon glyphicon-edit"></span></a>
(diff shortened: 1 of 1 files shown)


web/templates/user/_user_show_basedata.html
```diff
@@ -18,7 +18,7 @@
 
             {% call info_row("Name") %}
                 {% if current_user is privileged_for('user_change') %}
-                <a href="{{ url_for(".edit_user", user_id=user.id) }}">{{ user.name }} <span class="glyphicon glyphicon-edit"></span></a>
+                <span>{{ user.name }} </span><a href="{{ url_for(".edit_user", user_id=user.id) }}"><span class="glyphicon glyphicon-edit"></span></a>
```
marcelb98: I think the `<span>`s aren't needed.

Paktosan: Was unsure but apparently it does not make a difference so I removed the `<spans>`s.

Issue #288, opened 2018-11-12
Catch error messages


gehinz:
Cryptic sql exceptions are annoying.
Extract important information (at least for known errors) and present them in a more human friendly way.
Known errors that should be dealt with this way:

- [x] MAC address already exists in System (when editing an interface)

lukasjuhrich:
@gehinz are there still messages that annoy you, or can this issue be closed?

lukasjuhrich:
Given a lack of reaction and that currently we catch most things in most cases, I would like to close this.

Issue #240, opened 2018-09-10
Improve error handling when creating or moving users


Georgeto:
- [ ] Detect if a room has no patch port assigned and therefore no associated subnet can be found, instead of letting `get_free_ip` raise a `SubnetFullException`, because obviously no subnet means no free ip.
- [ ] Catch the exceptions that are thrown by the library.
- [ ] Maybe rework/rething error handling in the frontend in general?

lukasjuhrich:
> Maybe rework/rething error handling in the frontend in general?

I suggest starting in the backend with a proper base exception; we're very inconsistent at the moment with how we handle things.

PR #319, opened 2019-05-24
Add tasks, refactor hosts+interfaces, improve error handling

### LINKED ISSUE #62: Task scheduling
Pycroft should be able to execute asynchronous tasks at a given time.

This is not necessary for suspending a membership thanks to the intervalled membership logic, but for things like changing the mac address or the room a user lives in, schedule execution at a given time would drastically improve things for our members.

ibot3:
This PR adds and improves some things:

**Scheduled Tasks** 7445756
Tasks like moving, moving in, moving out a user can be scheduled to a specific date.
This is done with Celery. I introduced a new docker service, the "celery-worker".
Closes #62 

**(UI) Merging the interfaces table into the hosts table** c116ab7
The table to show the interfaces of a host is displayed as a subtable when clicking on a host row.
Further, all depending web functions for hosts and interfaces were moved to a new package called "host".
**This caused many "new" lines which aren't new but were just moved. Mostly from user/\_\_init\_\_.py to host/\_\_init\_\_.py**

**Error handling improvements** c71c47d
To avoid 5xx errors for users, all exceptions will be catched and an corresponding error message will be displayed. This isn't yet implemented everywhere.
References #288 and #240


pycroft/lib/task.py
```diff
@@ -0,0 +1,266 @@
...
+        task_impl = task_type_to_impl.get(task.type)()
+
+        try:
+            task_impl.execute(task)
+        except Exception as e:
+            task_impl.errors.append(str(e))
+
+        repair_session()
+
+        if task_impl.new_status:
```
Georgeto: Shouldn't this be `if task_impl.new_status is not None:`?

web/blueprints/helpers/exception.py
```diff
@@ -0,0 +1,53 @@
...
+        # Special case: Username already taken in Abe
+        # Should be removed after migration
+        if e.message == "Username already taken in Abe!":
+            flash("Dieser Benutzername wird bereits in Abe verwendet.", 'error')
+
+            session.session.rollback()
+        else:
+            raise e
+    except Exception as e:
+        traceback.print_exc()
```
ibot3: I am note sure if this actually works, should be tested.

pycroft/model/task.py
```diff
@@ -0,0 +1,91 @@
...
+    parameters_json = Column('parameters', JSONB, nullable=False)
+    created = Column(DateTimeTz, nullable=False)
+    creator = relationship('User')
+    creator_id = Column(Integer, ForeignKey("user.id"), nullable=False)
+    status = Column(Enum(TaskStatus), nullable=False, default=TaskStatus.OPEN)
+    errors = Column(JSONB, nullable=True)
+
+    @hybrid_property
+    def parameters(self):
+        from pycroft.lib.task import task_type_to_impl
```
Georgeto: The model should not import stuff from the library.

pycroft/model/task.py
```diff
@@ -0,0 +1,91 @@
...
+            raise ValueError("cannot find schema for task type")
+
+        parameters_schema = task_type_to_impl[self.type].schema()
+
+        return parameters_schema.load(self.parameters_json)
+
+
+    @parameters.setter
+    def parameters(self, _parameters):
+        from pycroft.lib.task import task_type_to_impl
```
Georgeto: The model should not import stuff from the library.

PR #320, opened 2019-06-01
Add option to confirm transactions made by non-treasurers

### LINKED ISSUE #318: Add page that shows transactions created by non-treasurers
There should be a page that shows all transactions that were created by non-treasurers. This should allow the treasurers to check transactions that were f.e. created by supporters when a bank account activity was required to be matched manually.

ibot3:
Implements and closes #318.
Also improves some minor stuff related to finances.


web/resources/js/table.js
```diff
@@ -99,6 +113,7 @@ export function btnFormatter(value, row, index) {
                 'btn_class': value['btn_class'],
                 'glyphicons': value['icon'],
                 'tooltip': value['tooltip'],
+                'target': target
```
Georgeto: Trailing comma would be nice.

web/resources/js/table.js
```diff
@@ -107,6 +122,7 @@ export function btnFormatter(value, row, index) {
                 'btn_class': value['btn_class'],
                 'glyphicon': value['icon'],
                 'tooltip': value['tooltip'],
+                'target': target
```
Georgeto: Trailing comma would be nice.

web/resources/js/table.js
```diff
@@ -115,6 +131,7 @@ export function btnFormatter(value, row, index) {
             'title': value['title'],
             'btn_class': value['btn_class'],
             'tooltip': value['tooltip'],
+                'target': target
```
Georgeto: Proper indentation and trailing comma would be nice.

web/blueprints/finance/__init__.py
```diff
@@ -608,6 +610,114 @@ def transactions_show_json(transaction_id):
...
+def transaction_delete(transaction_id):
+    transaction = Transaction.q.get(transaction_id)
+
+    if transaction is None:
+        flash(u"Transaktion existiert nicht.", 'error')
+        abort(404)
+
+    if transaction.confirmed:
+        flash(u"Diese Transaktion wurde bereits bestätigt und kann daher nicht gelöscht werden.", 'error')
+        abort(404)
```
Georgeto: Also applies to the other 404 added in this file. How does the user perceive the 404 error?
Is he redirected to the unconfirmed transactions overview by ` 'cancel_to': url_for('.transactions_unconfirmed')`?

web/blueprints/finance/__init__.py
```diff
@@ -348,10 +348,16 @@ def bank_account_activities_edit(activity_id):
             ).one()
             credit_account = activity.bank_account.account
 
+            confirmed = False
+
+            if current_user.member_of(config.treasurer_group):
+                confirmed = True
+
```
Georgeto: I would prefer `confirmed = current_user.member_of(config.treasurer_group)`.

web/blueprints/finance/__init__.py
ibot3: This and other 4xx errors are very rare and are mostly only triggered when the url is changed manually. So there dont't has to be a fancy redirection. It just displays an error like here:

https://pycroft.agdsn.de/user/1111111/

The user can then just use the back button of its browser.

However, 404 is not the correct status code for this.

web/blueprints/finance/__init__.py
```diff
@@ -601,13 +627,121 @@ def transactions_show_json(transaction_id):
...
+def transaction_confirm(transaction_id):
+    transaction = Transaction.q.get(transaction_id)
+
+    if transaction is None:
+        flash(u"Transaktion existiert nicht.", 'error')
+        abort(404)
+
+    if transaction.confirmed:
+        flash(u"Diese Transaktion wurde bereits bestätigt.", 'error')
+        abort(404)
```
Georgeto: Also not the correct status code here.

PR #321, opened 2019-06-01
Improve property group permissions

### LINKED ISSUE #203: Improve permission system
Currently, all users with the `groups_change_membership` property can add/remove all group memberships of all users.
It should not be possible to gather more permissions through this practice.
An option would be to invent a `protected` flag for groups, which memberships can then only be started/ended from root users.
Or the same flag for properties, where you can only begin/end memberships that affect this property, if you have the property yourself.

ibot3:
Closes #203 
Also adds the option edit property groups and an extra confirmation dialog before deleting a property group.


pycroft/lib/membership.py
```diff
@@ -120,4 +130,21 @@ def remove_member_of(user, group, processor, during=UnboundedInterval):
...
+              processor)
+
+    group.name = name
+    group.permission_level = permission_level
+
+@with_transaction
+def delete_property_group(group, processor):
+    log_event("Deleted property group '{}'.".format(group.name),
+              processor)
+
```
Georgeto: Shouldn't there be a permission_level check here?
`processor.permission_level  >= group.permission_level`

pycroft/lib/membership.py
```diff
@@ -120,4 +130,21 @@ def remove_member_of(user, group, processor, during=UnboundedInterval):
...
\ No newline at end of file
+                   user=user, author=processor)
+
+
+@with_transaction
+def edit_property_group(group, name, permission_level, processor):
+    log_event("Edited property group {} -> {}.".format((group.name, group.permission_level),
+                                                      (name, permission_level)),
+              processor)
+
```
Georgeto: Shouldn't there be a permission_level check here?
`processor.permission_level  >= group.permission_level`

web/blueprints/properties/__init__.py
```diff
@@ -54,7 +56,8 @@ def property_groups():
 def property_group_create():
     form = PropertyGroupForm()
     if form.validate_on_submit():
-        group = PropertyGroup(name=form.name.data)
+        group = PropertyGroup(name=form.name.data,
```
Georgeto: Although not an immediate security issue: Should it be allowed to create a property group with a permission level greater than your own?

ibot3: Regarding your three comments:
I added a check for creating/editing/deleting groups in the first place, but then realized that you could just create a group on your permission level with still all properties enabled that you want.
So it would have been needed a permission classification for properties too. I then decided that "groups_change" is kind of a super permission.
However, if wanted, I could also add a classification for properties.

Georgeto: Oops I missed the added `@access.require('groups_change')`, and the diff conveniently hid the already existing one on `property_group_create`, sorry. With that knowledge the whole changeset makes sense.

PR #322, opened 2019-06-16
Design improvements

lukasjuhrich:
- overhaul of the “overview” section, the most important states are communicated via icons
- removed the strong color of the `add an entry`-kind of table tool buttons (they're not a primary action)
- Adapt the colors of the `current_properties` labels
- add fancy hovering magic conveying information about which membership was the cause of granting / denying a property

### DIFF
--- a/web/resources/js/table.js
+++ b/web/resources/js/table.js
@@ -213,11 +213,18 @@ export function financeRowFormatter(row, index) {
     }
 }
 
-export function membershipRowFormatter(row, index) {
+export function membershipRowAttributes(row, index) {
     return {
         'data-row-grants': row.grants.join(' '),
         'data-row-denies': row.denies.join(' '),
     };
+}
+
+export function membershipRowFormatter(row, index) {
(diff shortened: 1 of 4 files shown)


Paktosan:
While I do think that the new design on the user page is quite refreshing, I think that the old layout is more easily read.
However I also have issues with my dev instance of Pycroft so I was unable to review everything yet.

web/resources/js/table.js
```diff
@@ -204,13 +204,87 @@ export function financeRowFormatter(row, index) {
...
+    }
+    return {};
+}
+
+$(function() {
+    let userPropSel = $('.userprop[data-property-name]');
+    userPropSel.mouseenter(function(ev) {
+        let propitem = ev.currentTarget;
+        let propname = propitem.attributes['data-property-name'].value;
+        console.log("hovered " + propname);
```
ibot3: Should these console prints be kept in the productive version?

PR #336, opened 2019-09-29
create wifi passwords for users

marcelb98:
I've introduced a new column for all users with a new password which is used for wifi access. Per default this password is not set. On the users detail page is a new button to create or reset the wifi password. When this is done a new "wifi usersheet" is generated.


pycroft/lib/user.py
```diff
@@ -133,15 +134,18 @@ def store_user_sheet(new_user, plain_password, timeout=15, generation_purpose=''
     :param str plain_password:
     :param str generation_purpose:
     :param int timeout: The lifetime in minutes
+    :param bool wifi: Generate a user sheet for wifi access instead of the generic one
     """
-    pdf_data = b64encode(generate_user_sheet(new_user, plain_password, generation_purpose)).decode('ascii')
+    if wifi is True:
+        pdf_data = b64encode(generate_wifi_user_sheet(new_user, plain_password)).decode('ascii')
+    else:
+        pdf_data = b64encode(generate_user_sheet(new_user, plain_password, generation_purpose)).decode('ascii')
```
lukasjuhrich: For the future: Coud've been
```python
user_sheet = (
    generate_wifi_user_sheet(new_user, plain_password) if wifi
    generate_user_sheet(new_user, plain_password, generation_purpose)
)
pdf_data = b64encode(user_sheet).decode('ascii')
```

pycroft/lib/user.py
```diff
@@ -133,15 +134,18 @@ def store_user_sheet(new_user, plain_password, timeout=15, generation_purpose=''
     :param str plain_password:
     :param str generation_purpose:
     :param int timeout: The lifetime in minutes
+    :param bool wifi: Generate a user sheet for wifi access instead of the generic one
     """
-    pdf_data = b64encode(generate_user_sheet(new_user, plain_password, generation_purpose)).decode('ascii')
+    if wifi is True:
```
lukasjuhrich: `is True` is an antipattern, because this breaks python's duck-typing. Correct would be just `if wifi` – then, one could pass other values (such as a mock) which evaluate to `True` under the lense of `__bool__`, `__len__`, etc.

cf
- [__bool__ (python3 docs)](https://docs.python.org/3/reference/datamodel.html#object.__bool__)
- [PEP8, Programming recommendations, 21st point](https://www.python.org/dev/peps/pep-0008/#programming-recommendations)

pycroft/helpers/printing/__init__.py
```diff
@@ -270,6 +270,181 @@ def generate_user_sheet(user, user_id, plain_password, generation_purpose=''):
...
+    im_fb = Image(ASSETS_FACEBOOK_FILENAME, 0.4 * cm, 0.4 * cm)
+    im_t = Image(ASSETS_TWITTER_FILENAME, 0.4 * cm, 0.4 * cm)
+    im_logo = Image(ASSETS_LOGO_FILENAME, 3.472 * cm, 1 * cm)
+
+    if user.room:
+        shortinfo = Paragraph('{dorm}<br/>{name}<br/>{level}/{room}'.format(
+            dorm=str(user.room.building.short_name),
+            name=user.name,
+            level=str(user.room.level),
+            room=str(user.room.number)
```
lukasjuhrich: The `str()`s are unnecessary, because `format()`/[`__format__`](https://docs.python.org/3/reference/datamodel.html#object.__format__) does that for you.

web/templates/user/user_reset_wifi_password.html
```diff
@@ -0,0 +1,18 @@
+{#
+ Copyright (c) 2017 The Pycroft Authors. See the AUTHORS file.
```
lukasjuhrich: Oh, is it 2017 again? :D

Issue #67, opened 2016-12-03
User Deletion


lukasjuhrich:
We should think about a procedure to delete users. The unix_account might be considered independent of this question.

lukasjuhrich:
This touches the following behaviors:

## Use Case: User moves out continuing use of services
- The room reference will be removed
- We need to know the _current_ address after the dormitory for that.

## Use Case: User moves out without continuing use of services having nonzero balance
- We need to know the _last known_ address for the finance information (de facto the former `room`s address) → Just leave what's in `user.address`
- The splits still need to be associated to some user, so we desire archiving over deletion
- **Open Question:** Do we want to retain group memberships to know what the actual periods of formal AGDSN membership were? What about `UserLogEntries`?

When moving out, the following follow-up action should be considered:

## Use Case: Closing of finance account
- A user without a room or services and with a zero balance in his account _could be deleted_
- The account would still exist, so we still retain the transactions while discarding unneccessary personal information.

## Use Case: limitation of claim („Verjährung“)
- A user with a nonzero finance balance _could be deleted_ as well after three years due to limitation of claim.
- **Open Question:** It's still not clear how we want to deal with this regarding the financial process.

## Unclear: How to deal with former admins
A lot of management entities like logentries or an `MT940Error` reference a `processor`. It is unclear how that should be handled when the former admin e.g. demands deletion of his/her data.

## Possible realization of „archiving“
One could consider splitting up an `ArchivedUser` from a `User` both inheriting from an `AbstractUser` (RIP naming) following the following criteria:
- a `User` is characterized by the ability to live in a room, being referenced for `UserTask`s, requiring a `UnixAccount`, and having `Host`s and all the technical stuff to be introduced in #330 #331  #73 etc.
- an `AbstractUser` is characterized by the ability to have a finance_account, by requiring a name, and requiring an address. As mentioned above, possibly also by the ability to be referenced by `Membership`s and `LogEntries`
- An `ArchivedUser` is characterized by the ability to… _optionally_ have a unix_account? At least nothing of importance here.

Depending on the technical details, we could possibly drop the sibling construction and opt for a parent-child construction `ProFormaUser←User` – „archiving“ would the mean dropping the polymorphic `User` identity of a `ProFormaUser`.

lukasjuhrich:
We want the following three-step model:
1. Marking a user for deletion by adding them to a „to be deleted“ group, which can deny all properties
[...]

PR #338, opened 2019-10-03
Generalized location

### LINKED ISSUE #305: Generalize "Wohnheim" to location
Currently, it is impossible to manage external residences in Pycroft. However, storing this data is required by law. The "Wohnheim"/building system could be generalized so that they represent locations, then a new location Dresden can be created having addresses rather than room numbers.

lukasjuhrich:
* [14/16] Battle plan

- [X] Create Address Table
- [X] Create FKeys User/Room → Table
- [X] Create =Config.dummy_address=
- [X] Observe tests failing and their failure reasons, fix fixtures
- [X] Test that =move_in= sets the address of the room
- [X] Set address in =move_in=
- [X] Test =move= changes address if =user.room.address == user.address=
  + For that, add property =User.has_custom_address=
- [X] Test =move= does _not_ change address if user has different address than room
- [X] Implement the above behavior
- [X] Test moving out sets =config.dummy_address= if user doesn't have custom address
- [X] Test moving out does not change =config.dummy_address= if user has a custom address
- [X] Implement behavior in =move_out=
- [X] Schema Migrations (Table, FKeys, Config)
- [X] Data Migrations (User, rooms) – Note that we not always get a
  _new_ address for each room.  Non-inhabitable rooms should just
  refer to a building-level address!

  1. Look at other migrations
  2. Add dummy room
  3. Set =config.dummy_room=
  4. Add addresses for Building (perhaps look store the mapping
     building→address in a temporary table)
  5. Add addresses for rooms if inhabitable, else set to building-level address
  6. Set Address for users without a room
  7. Set address to rooms address for users with a room
  8. Ta-daa!
- [x] Change behavior of `move_out` (leave address instead of changing to `dummy_address` ⇒ remove `config.dummy_address_id`)
- [x] Expose info in UI (room & user overview)
  + Test case for string formatting
  + User overview (hover question mark icon?)
  + Room overview
- [x] Create user without room, but with address, and check datasheet
  capabilities.
- [x] ~Expand user.lib test cases (perhaps in a different issue / pr)~ see #340

This PR fixes #305


pycroft/model/alembic/versions/ff700db83195_add_address_relation.py
```diff
@@ -0,0 +1,244 @@
...
+
+def add_dummy_address(session):
+    # noinspection SpellCheckingInspection
+    dummy_addr = _Address(street="Niemandsstraße", number="42", zip_code="33602",
+                          city="Bielefeld", state="Niedersachsen")
+    session.add(dummy_addr)
+    session.commit()
+    session.refresh(dummy_addr)
+    dummy_id = dummy_addr.id
+    return dummy_id
```
ibot3: Why are we using a dummy address instead of a nullable address_id?

ibot3: And what is the backround of always updating the user `address_id` on room changes instead of keeping it on `null`? The `address` attribute could be a hybrid property returning `address_id` if it is not `null`, `room.address_id` if the user is living in a room or `null` if both conditions weren't met.
The current implementation could lead to inconsistency in the database.

lukasjuhrich: See [this comment](https://github.com/agdsn/pycroft/issues/305#issuecomment-537691101).

There's a few questions/decisions which are relevant here.

## When to enforce that every user has an address

At some point in the future, to conform with TKG requirements, we should ensure that **every user has a fully qualified address**. I think it's simpler to already represent this convention in the schema and at a later point in time fix the data, instead of now not requiring the constraint. Because if we add the constraint later, we will have a migration that only works when e.g. you don't have any users without an address information. I think that's not a good idea from an operational perspective.
The cleanup work of deleting the users (see #67) must be done anyway, and whether these users have a `null` address or the dummy address does not matter from a data perspective.

## How to enforce that every user has an address
To ensure this on the sql level, we could 
- Make `user.address_id` non-nullable
- Create a check constraint verifying that every `user` has either an `address_id` or a `room_id`.

Following [@mmarx's comment](https://github.com/agdsn/pycroft/issues/305#issuecomment-469382507), I opted for the first solution due to its simplicity.

## Is the potential inconsistency bad?
For a weak argument, it might even be a use case that someone wants to have a different fully-qualified address than where his room is located.

For a stronger argument, our schema isn't perfectly irredundant, anyway, and relies on consistency by convention because the library functions act „correctly“. For instance, the schema allows a user to have a host without an interface. There's no use case for that.
Explicitly setting an address in just two instances is imo quite a low cost for the simplicity.

For a conceptual argument: The `room` as we use it is actually more of a technical entity: we  use it

- to know which switch ports (also patch ports) are associated to the user (to establish access and diagnostics)
- after that, potentially to identify a person by their room in a support scenario.
This is not necessarily coupled to postal identification.

So i'd argue that from that perspective, it's not incorrect to have the postal information (`user.address_id`) decoupled from the technical location (`user.room_id`) – on the contrary.

lukasjuhrich:
BTW, The datasheet has been adapted as follows:
- It looks the same if the user's address is the same as his room's address
- Else, it looks like this:
pycroft_datasheet_address

pycroft/model/config.py
```diff
@@ -3,6 +3,8 @@
 # the Apache License, Version 2.0. See the LICENSE file for details.
 from sqlalchemy import CheckConstraint, Column, ForeignKey, Integer
 from sqlalchemy.orm import relationship
+
+from pycroft.model.address import Address
```
ibot3: Probably unnecessary import

lukasjuhrich: Good catch, there was one in the `ConfigFactory` as well.

PR #346, opened 2019-10-13
Address fix

### LINKED ISSUE #343: Addresses without `addition` column may not be unique
In contrast to the unique constraint spanning all columns, we need additional partial unique constraints which don't take into account nullable columns when they are null.

See e.g. https://dba.stackexchange.com/a/9760/144999

lukasjuhrich:
Fixes #343


pycroft/model/alembic/versions/ce8c8cebcbf4_make_address_state_and_addition_not_null.py
```diff
@@ -0,0 +1,37 @@
...
+
+"""
+from alembic import op
+import sqlalchemy as sa
+import pycroft
+
+
+# revision identifiers, used by Alembic.
+revision = 'ce8c8cebcbf4'
+down_revision = '38fa5154b920'
```
ibot3: Must be `5b6f5a33e426`

pycroft/model/alembic/versions/ce8c8cebcbf4_make_address_state_and_addition_not_null.py
```diff
@@ -0,0 +1,37 @@
+"""Make address.state and addition not null
+
+Revision ID: ce8c8cebcbf4
+Revises: 38fa5154b920
```
ibot3: Must be `5b6f5a33e426`

PR #353, opened 2019-11-22
manual mt940 bank import

marcelb98:
This PR adds the possibility to import transactions without the use of FinTS.
The transactions need to be downloaded from the online banking as a MT940 file which will be uploaded to Pycroft. The MT940 is saved like an unsuccessfull FinTS import. So, after uploading the file, the transactions can be reviewed and imported.

### DIFF
--- a/web/blueprints/finance/__init__.py
+++ b/web/blueprints/finance/__init__.py
@@ -46,7 +46,8 @@ from web.blueprints.finance.forms import (
     AccountCreateForm, BankAccountCreateForm, BankAccountActivityEditForm,
     BankAccountActivitiesImportForm, TransactionCreateForm,
     MembershipFeeCreateForm, MembershipFeeEditForm, FeeApplyForm,
-    HandlePaymentsInDefaultForm, FixMT940Form, BankAccountActivityReadForm)
+    HandlePaymentsInDefaultForm, FixMT940Form, BankAccountActivityReadForm,
+    BankAccountActivitiesImportManualForm)
 from web.blueprints.finance.tables import FinanceTable, FinanceTableSplitted, \
     MembershipFeeTable, UsersDueTable, BankAccountTable, \
     BankAccountActivityTable, TransactionTable, ImportErrorTable, \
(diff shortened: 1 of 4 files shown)


web/blueprints/finance/forms.py
```diff
@@ -116,6 +116,11 @@ class BankAccountActivitiesImportForm(Form):
     do_import = BooleanField(u"Import durchführen", default=False)
 
 
+class BankAccountActivitiesImportManualForm(Form):
+    account = SelectField(u"Bankkonto", coerce=int)
```
ibot3: Use a `QuerySelectField` here instead.
You can find several examples in `web/blueprints/user/forms.py` and the documentation at http://wtforms.simplecodes.com/docs/0.6/ext.html#module-wtforms.ext.sqlalchemy.fields.

This loads the choices asynchronously and integrates better with SQLAlchemy.
`form.account.data` will return a BankAccount directly.
The `get_label` argument can be a callable to display the correct values in the select field.
`form.account.choices = ...` is not required anymore.

web/blueprints/finance/__init__.py
```diff
@@ -235,6 +236,31 @@ def bank_accounts_import():
                            transactions=transactions,
                            old_transactions=old_transactions)
 
+@bp.route('/bank-accounts/importmanual', methods=['GET', 'POST'])
+@access.require('finance_change')
+def bank_accounts_import_manual():
+    form = BankAccountActivitiesImportManualForm()
+    form.account.choices = [(acc.id, acc.name) for acc in BankAccount.q.all()]
```
ibot3: Not needed with a QuerySelectField (see other comment).

web/blueprints/finance/__init__.py
```diff
@@ -235,6 +236,31 @@ def bank_accounts_import():
...
+
+    if form.validate_on_submit():
+        bank_account = BankAccount.q.get(form.account.data)
```
ibot3: Can be `bank_account = form.account.data` when using a QuerySelectField (see other comment).

Issue #375, opened 2020-05-15
Flickercode support / TAN input


Paktosan:
Since PSD 2, reading requests via FinTS may require a TAN to complete so we require a way to enter it in Pycroft. The flow would be the following:
1. User starts action involving FinTS
2. If TAN is required goto 3, else 5
3. User gets redirected to page showing Flickercode and input for TAN
4. If TAN is correct goto 5, else goto 3 and show message to try again
5. Action completes

For this to work we require some kind of session storage so we can save some data of FinTS library while the user enters the TAN.

lukasjuhrich:
What is the current workflow? As far as I know, the flickercodes have been deprecated in favor of QR codes, correct?

ibot3:
I think this can be closed.
There is no TAN required when the online banking login has been unlocked once within the last (60?) days.
So the workflow is to login to the online banking portal if a TAN is required.

lukasjuhrich:
Closing this and PR #376, consequently.

PR #376, opened 2020-05-15
[WIP] FinTS TAN input

Paktosan:
Also see #375


lukasjuhrich:
If there's time after the address review and deployment, we can look at this tomorrow.

This should definitely not be delivered via in-site JS, but as a proper module (perhaps even TS), so that webpack can go into action.

It would help adding a small to-do list so it's transparent what still needs to be done.

lukasjuhrich:
Closing as discussed in #375

PR #382, opened 2020-06-07
Membership fee per dorm

### LINKED ISSUE #197: Add integration tests for membership fee booking
This assignment was taken out of the requirements for the #179 merge and should be added at some point in the future.

### LINKED ISSUE #328: Book membership fees seperated by dormitory
The membership fees should be booked to a separate finance account per dormitory. This requires some changes.
When booking a fee, all fee accounts need to be checked for an already existent transaction. 
There currently is a problem that the users room is only one value, independent of movements. If a user moves out, there is no room set anymore and it can not be determined to which fee account the transaction should go.

### LINKED ISSUE #367: Tabbed design for user page
The user page (user_show) may could be better to use if we would use a tabbed layout instead a long page with multiple sections.
I would propose that the base data is always visible and underneath the tabs for Hosts, Logs, Memberships, ...

ibot3:
This PR adds the ability to book membership fees separated by dormitory (closes #328).
An additional room history is created for every user that keeps track in which period a user lived in a room.
If a fee is booked, it is checked in which room a user lived at the end of the fee period and at the beginning of the period as an alternative.
By default, the current fee account is used and the account of a dormitory can be changed in the database.

This PR also add test cases for fee booking and handling of overdue users. (closes #197).

I also changed the design of the user overview to a tabbed layout (closes #367).

### DIFF
--- a/pycroft/model/finance.py
+++ b/pycroft/model/finance.py
@@ -192,7 +192,7 @@ class Split(IntegerIdModel):
     transaction = relationship(Transaction,
                                backref=backref("splits",
                                                cascade="all, delete-orphan"))
-    __table_args = (
+    __table_args__ = (
         UniqueConstraint(transaction_id, account_id),
     )
 
@@ -343,7 +343,7 @@ class BankAccountActivity(IntegerIdModel):
                          backref=backref("bank_account_activity",
                                          uselist=False))
 
-    __table_args = (
+    __table_args__ = (
         ForeignKeyConstraint((transaction_id, account_id),
                              (Split.transaction_id, Split.account_id),
                              onupdate='CASCADE',
--- a/pycroft/model/finance.py
+++ b/pycroft/model/finance.py
@@ -213,9 +213,9 @@ manager.add_function(
           ELSE
             s := COALESCE(NEW, OLD);
           END IF;
-          
+
           SELECT COUNT(*) = 0 INTO transaction_deleted FROM "transaction" WHERE "id" = s.transaction_id;
-            
+
           SELECT COUNT(*), SUM(amount) INTO STRICT count, balance FROM split
               WHERE transaction_id = s.transaction_id;
           IF count < 2 AND NOT transaction_deleted THEN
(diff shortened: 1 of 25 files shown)


pycroft/model/finance.py
```diff
@@ -192,7 +192,7 @@ class Split(IntegerIdModel):
     transaction = relationship(Transaction,
                                backref=backref("splits",
                                                cascade="all, delete-orphan"))
-    __table_args = (
+    __table_args__ = (
```
lukasjuhrich, line 195: Oh no. Thanks for fixing that… :/

pycroft/model/finance.py
```diff
@@ -51,7 +51,7 @@ def __contains__(self, date):
         return date in closed(self.begins_on, self.ends_on)
 
     __table_args__ = (
-        CheckConstraint('begins_on < ends_on'),
+        CheckConstraint('begins_on <= ends_on'),
```
lukasjuhrich, line 54: I too find this sensible. Just for clarification, do we have a use case for one-point-intervals?

pycroft/model/user.py
```diff
@@ -341,6 +344,46 @@ def permission_level(self):
...
+        volatility='volatile', strict=True, language='plpgsql'
+    )
+)
+
+manager.add_trigger(
+    User.__table__,
+    ddl.Trigger(
+        'user_room_change_update_history_trigger',
+        User.__table__,
+        ('UPDATE', 'INSERT'),
```
lukasjuhrich, line 381: What happens on room deletion (the only thing taking action should be the cascades)?
I do realize we're unlikely to ever do that.

tests/__init__.py
```diff
@@ -154,13 +156,14 @@ class FixtureDataTestBase(SQLAlchemyTestCase, DataTestCase, unittest.TestCase):
     """
     @classmethod
     def setUpClass(cls):
-        SQLAlchemyTestCase.setUpClass()
+        super().setUpClass()
```
lukasjuhrich, line 159: :+1:

tests/factories/property.py
```diff
@@ -3,6 +3,7 @@
 
 import factory
 
+from pycroft.model import session
```
lukasjuhrich, line 6: unused, fixing that in my branch.

PR #427, opened 2021-02-10
Address creation

### LINKED ISSUE #344: Room cannot be added
The creating a room in the frontend does not work because no address is set.

lukasjuhrich:
This PR
- Introduces typescript
- Creates a generic `typeahead` component
- Fixes #344 by introducing `get_or_create_address` and proper form inputs for the `create_room` endpoint.

The room creation has two key features:
- All fields except for `addition` are pre-filled by these values that have the highest occurrence among the addresses of the other rooms in the respective buildings
- There is an simple autocompletion for each field.


web/resources/js/hello.ts
```diff
@@ -0,0 +1,7 @@
+import $ from "jquery";
```
georgeto, line 1: Remove this commit from the PR? With `generic-typeahead.ts` there seems to be a proper typescript module now.

web/resources/js/hello.ts
```diff
@@ -5,3 +5,14 @@ document.addEventListener('DOMContentLoaded', () => {console.log("This is runnin
 $(function () {
     $('[data-toggle="tooltip"]').tooltip()
 });
+
```
georgeto, line 8: Remove this commit from the PR? With `generic-typeahead.ts` there seems to be a proper typescript module now.

ibot3:
For me, the `test_address_unique` test is failing. Can you confirm that?

lukasjuhrich:
> For me, the `test_address_unique` test is failing. Can you confirm that?

Yes. Running the complete suite again and fixing tests afterwards.

lukasjuhrich:
Fixed, thanks

Issue #432, opened 2021-02-11
Clean up unused addresses in the model when updating / deleting reference to it


lukasjuhrich:
We do not clean up after orphaned addresses. This is not likely to cause a problem, but unnecessarily lazy.

Issue #428, opened 2021-02-10
Regression: allow creation of users without a room


ibot3:
It should be possible to edit the address of a user and the address of a room.

Furthermore, creating a user without room is currently not possible, because no address is set. We could either use a dummy address or add the address fields to the user create form.

lukasjuhrich:
Refs #344

lukasjuhrich:
Are membership requests without a `room` relevant to us in any way?
`create_member_request` does not enforce that a `room` is set.
If I understand correctly such requests could never have been parsed since there is no way for an address to be specified.

The Background is that I would like to avoid adding even more cludder (read: 5+ input forms for the address) to our generic user creation form, and instead split things up into the two use cases
1. Creating a new user in a known room (perhaps even removing this largely useless membership select field)
2. Creating an external user, requiring specification of an address.

lukasjuhrich:
Also, why is the email [suddenly required](https://github.com/agdsn/pycroft/blob/7005136b68038907a22fdca101faf012df4bd6c2/web/blueprints/user/forms.py#L116) when editing members, but `PreMember.email` is nullable? What's the right constraint here?

ibot3:
The email should be required for a member request. If that's not enforced by pycroft, it is probably a mistake.

Member requests can have the room set to `None` for some time. However, for transforming a member request to a user, a room is required.

lukasjuhrich:
Fixed by #433

Issue #408, opened 2020-10-06
Room edit should force stating the new address


lukasjuhrich:
Since addresses are now decoupled from the room's data, editing the number of a room should enforce providing a new address.
This should affect the UI as well as the signature of `room_edit` (?) in the lib.

I would consider this a (usability) bug because the current process leads to inconsistent addresses if the user is not _specifically_ aware of this behavior.

lukasjuhrich:
Refs #344

Issue #347, opened 2019-10-13
Make user's address editable


ibot3:
It should be possible to edit a users address in the UI.

With #338, the option to create a user without moving the user into a room has been removed. This should be possible again. Users without a room could get a dummy address by default.

lukasjuhrich:
Plan:
- [ ] Add `edit_address` function to `lib`
- [ ] Add tests for all the scenarios (with / without room / previous address set)
- [ ] Implement the lib functionality
- [ ] Allow for address editing in UI, and ensure addresses are represented whenever they differ from a user's room

PR #433, opened 2021-02-12
Address editing

lukasjuhrich:
- [X] #432
- [x] #347
- [x] #408
- [x] #428


lukasjuhrich:
Der restliche Mist sollte wenigstens wesentlich schneller gehen.
Aber sicherlich gute Vorbereitung auf #330, wo auch ne Menge auf Postgres-Ebene passieren muss.

lukasjuhrich:
Only #408 is missing and that shouldn't be too much work.
Opening this issue for review, since a lot of cleanup happened along the way.

web/blueprints/user/forms.py
```diff
@@ -42,6 +48,93 @@ def validate_unique_login(form, field):
...
+            level = getattr(form, self.level_field).data
+            building = getattr(form, self.building_field).data
+        except AttributeError:
+            return
+
+        return Room.q.filter_by(number=number, level=level, building=building).one_or_none()
+
+    def similar_users(self, our_name, room: Room):
+        return [u for u in room.users
+                if SequenceMatcher(None, our_name, u.name).ratio() > self.ratio]
```
ibot3, line 83: I had a point on my todo list to improve this check. I would recommend checking for the address or the room history of the users instead of just the current room. This is probably not difficult to implement.

web/templates/user/user_create.html
```diff
@@ -15,28 +15,6 @@
 {% endblock %}
 
 {% block page_script %}
-    {{ resources.link_script_file('lazy-load-select.js' | require) }}
```
ibot3, line 18: The `lazy-load-select.js` is missing on this page.
I would propose to use the `generic_form` template for this form, where the script is already included.  Only the `confirmable-error.js` script must be added there.

web/templates/user/_user_show_basedata.html
```diff
@@ -43,6 +43,7 @@
                 <span class="fa fa-home fa fa-home"></span>
                 <span>{{ user.address }}</span>
               {% endif %}
+              <a href="{{ url_for('.edit_user_address', user_id=user.id) }}"><span class="fa fa-edit"></span></a>
             {% elif user.address is not none %}
               <span class="fa fa-home"></span>
               <span>{{ user.address }}</span>
```
ibot3, line 49: The button for editing the users address should also be visible if no room is set.

Also: The else condition can be removed, as address cannot be `None`.

ibot3 (changes requested):

web/templates/user/user_create.html
lukasjuhrich: I think instead I will do the „clean“ version and include all form-input-features in a `form.js` bundle, which can then be included. We should do something similar with regards to `table`-related scripts (bootstrap table etc), and remove these from the `main` bundle to reduce unneccessary initial load.

lukasjuhrich: For this PR, I will just do a quickfix tho and re-add these references to not introduce unneccessary diff.

web/templates/user/_user_show_basedata.html
lukasjuhrich: Fixed

web/blueprints/user/forms.py
lukasjuhrich: I'm confused by this statement. We are talking about the uniqueness check when e.g. creating a user; before the user is created, he does not have any room history.
The only info we have is „Somebody wants to create a user with name `our_name` and move them to `room`“.

lukasjuhrich: cleared in manual conversation.

lukasjuhrich: Fixed in 71e5905eadd43ad9c503287d1decff270f1af9e8

ibot3 (approved):

PR #462, opened 2021-04-21
Improvements in the Tasks implementation

lukasjuhrich:
- [x] Upgrade marshmallow
- [x] Deserialize to object instead of dict
- [x] Serialize from object instead of dict
- [x] Test move_out scheduling
- [x] Test move_out impl
- [x] Test move scheduling
- [x] Test move impl
- [x] Test move_in scheduling
- [x] Test move_in impl


pycroft/model/task_serialization.py
```diff
@@ -0,0 +1,66 @@
...
+    @post_load
+    def build(self, data, **kwargs):
+        return UserMoveOutParams(**data)
+
+
+@dataclass
+class UserMoveOutParams:
+    comment: str
+    end_membership: bool
+
```
ibot3, line 21: The `*Schema` and `*Params` classes cannot be combined with marshmallow? Because this seems to be a little bit unnecessary. [Pydantic](https://pydantic-docs.helpmanual.io/) for example uses the python types for validating and you only need one class per Schema.

ibot3 (approved):

pycroft/model/task_serialization.py
lukasjuhrich: Doesn't seem that way, but there is [marshmallow-dataclass](https://pypi.org/project/marshmallow-dataclass/).
I'd say we can think about this in the future, but in this refactoring PR I don't want to introduce a new library.

lukasjuhrich: It's also not that bad because the outer code does not need to use both the class and the schema: Once `Task.params` gets assigned a value, the required schema is inferred from the `TaskType` and dumping is initiated.
The „end-user“ only needs to import e.g. `UserMoveParams`, and not `UserMoveSchema`.
Resolving.

ibot3 (approved):
Looks good! Thank you for tidying up the tasks and some tests :)

pycroft/lib/task.py
```diff
@@ -72,93 +70,90 @@ def schedule(self, due, user, parameters, processor):
...
-                self.errors.append(
-                    "Tried to move user, but target room did not exist.")
-        else:
-            self.errors.append(
-                "Tried to move in user, but user was already living in a dormitory.")
+        room = Room.q.filter_by(
+            number=parameters.room_number,
+            level=parameters.level,
+            building_id=parameters.building_id,
+        ).first()
```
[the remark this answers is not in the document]
lukasjuhrich: fixed in 0dc71f2eb

pycroft/lib/task.py
```diff
@@ -72,93 +70,90 @@ def schedule(self, due, user, parameters, processor):
...
+        if task.user.room is not None:
+            self.errors.append("Tried to move in user, "
+                               "but user was already living in a dormitory.")
+            return
+
+        room = Room.q.filter_by(
+            number=parameters.room_number,
+            level=parameters.level,
+            building_id=parameters.building_id,
+        ).first()
```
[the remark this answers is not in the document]
lukasjuhrich: Fixed in 0dc71f2eb

PR #473, opened 2021-05-23
Match team transactions

### LINKED ISSUE #354: Automatch bank transactions of teams
When  automatching transaction we could match transactions which contain the [YEAR]-[TEAM][NUMBER] scheme.

lucashecht:
This PR adds the ability to match transactions to their corresponding teams based on the case number in the reference, similar to the already existing user matching.

Regexes for the case numbers and Account-IDs of the corresponding teams need to be inserted in the table ```AccountPattern```.

Closes #354


pycroft/lib/finance.py
```diff
@@ -829,12 +829,11 @@ def build_transactions_query(account, search=None, sort_by='valid_on', sort_orde
 def match_activities():
     """Get a dict of all unmatched transactions and a user they should be matched with
 
-    :param BankAccount bank_account: The BankAccount to get the unmatched transactions from
-
     :returns: Dictionary with transaction and user
     :rtype: Dict
```
lukasjuhrich, line 833: This rtype now outdated and should be replaced by a proper type annotation.
However, using dicts to transport structured data is a bad idea anyway (not your fault).
The dicts should be replaced by `@dataclass` types.

This allows someone else to more easily because the IDE can help you with completion, and because there is a central locatiion where one can look up what data is _actually_ being passed (namely the type definition).

pycroft/model/finance.py
```diff
@@ -342,6 +346,11 @@ class BankAccountActivity(IntegerIdModel):
     split = relationship(Split, foreign_keys=(transaction_id, account_id),
                          backref=backref("bank_account_activity",
                                          uselist=False))
+    matching_patterns = relationship(
+        AccountPattern,
+        primaryjoin='foreign(BankAccountActivity.reference).op("~*", is_comparison=True)(remote(AccountPattern.pattern))',
```
lukasjuhrich, line 351: Line breaking this string would be appropriate

web/blueprints/finance/__init__.py
```diff
@@ -412,81 +411,106 @@ def bank_account_activities_edit(activity_id):
...
 def bank_account_activities_match():
-    FieldList = [
-        # ("Field-Name",BooleanField('Text')),
-    ]
+    matching_user, matching_team = match_activities()
+
+    field_list_user, matched_activities_user = _create_field_list_and_matched_activities_dict(matching_user,
+                                                                                              "user")
+    field_list_team, matched_activities_team = _create_field_list_and_matched_activities_dict(matching_team,
+                                                                                              "team")
```
lukasjuhrich, line 419: This would be one of the scenarios where an escape line-continuation would help:

```suggestion
    field_list_user, matched_activities_user \
        = _create_field_list_and_matched_activities_dict(matching_user, "user")
    field_list_team, matched_activities_team \
        = _create_field_list_and_matched_activities_dict(matching_team, "team")
```

Although a concise name would be even better.

web/blueprints/finance/__init__.py
```diff
@@ -412,81 +411,106 @@ def bank_account_activities_edit(activity_id):
...
+
+def _create_field_list(matching):
+    field_list = []
+    for activity, entity in matching.items():
+        field_list.append(
+            (str(activity.id), BooleanField('{} ({}€) -> {} ({})'.format(
+                activity.reference, activity.amount, entity.name, entity.id
             ))))
 
+    return field_list
```
lukasjuhrich, line 474: This would be a good use case for a generator function:
```suggestion
def iter_matching_fields(matching: dict):
    for activity, entity in matching.items():
        yield str(activity.id), \
            BooleanField('{} ({}€) -> {} ({})'.format(
                activity.reference, activity.amount, entity.name, entity.id
            ))
```

And then use it above with `field_list_user = list(iter_matching_fields(…)`

lukasjuhrich: Also, you pass all of `matching` even though you only need `matching.items()`!
A good metaphor for this principle is “you don't give someone a wallet for them to take out the money, you give someone the money they need”.

Not particularly relevant here, but a good principle to obey for building maintainable code.

lukasjuhrich:
FYI, I will fix most of the things I remarked myself, as I have the code open already.

lukasjuhrich:
- [x] Basic unit test works
- [x] Basic manual test works

image

Yay!

lukasjuhrich:
@lucashecht I added the patterns to the UI for transparency (not everyone can just quickly look inside the database ;) )

image

lucashecht:
Great! Thank you for your input and additions @lukasjuhrich!

PR #475, opened 2021-05-26
SQLAlchemy 1.4 / 2.0 compat

### LINKED ISSUE #471: Update to SQLA 1.4
This does not mean using [v2.0-style query syntax](https://docs.sqlalchemy.org/en/14/glossary.html#term-2.0-style), just upgrading and not breaking things.

lukasjuhrich:
This PR
- Upgrades sqlachemy to `~=1.4.15` and `alembic` to `~=1.6.4`
- Establishes full compatibility with SQLAlchemy v1.4 and makes it work
- Establishes full compatibility with SQLAlchemy-to-come v2.0 by virtue of resolving all the issues leading to `warning`s
- In virtue of the above drastically improves (by 10x!) the performance of the query determining which users have to pay (see the difference between [this old](https://git.agdsn.de/-/snippets/103) and [this new](https://git.agdsn.de/-/snippets/104) compiled query).
- Turns the `test` suite into a `future=True` engine

For more details, see the commit messages.

To do:
- [x] Run the tests and see them pass
- [ ] Actually use pycroft in light of all these changes

Otherwise, ready fo review.

Fixes #471


ibot3 (approved):
Looks good :)

PR #477, opened 2021-06-02
Add user to a `pre-member` group, if the move-in date is in the future

ibot3:
Add a new user to a `pre-member` group, if the move-in date is in the future.

Required, so that users that will be a member in the future are able to login on SIPA before their membership starts.

### DIFF
--- a/pycroft/lib/user.py
+++ b/pycroft/lib/user.py
@@ -375,9 +375,9 @@ def move_in(
         user.birthdate = birthdate
 
     if begin_membership:
-        if user.member_of(config.external_group):
-            remove_member_of(user, config.external_group, processor,
-                             closedopen(session.utcnow(), None))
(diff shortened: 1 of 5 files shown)


pycroft/lib/user.py
```diff
@@ -1156,6 +1156,10 @@ def finish_member_request(prm: PreMember, processor: Optional[User],
     message = deferred_gettext("Created from registration {}.").format(str(prm.id)).to_json()
     log_user_event(message, processor, user)
 
+    if move_in_datetime > session.utcnow():
+        make_member_of(user, config.pre_member_group, processor,
+                       closed(session.utcnow(), move_in_datetime + timedelta(hours=3)))
```
lukasjuhrich, line 1161: …why three hours in the future?

ibot3: Has been improved with newer commits.

PR #480, opened 2021-06-16
Upgrade to Bootstrap 5

### LINKED ISSUE #448: Update to Bootstrap 5
A few days ago, the [final beta](https://blog.getbootstrap.com/2021/03/23/bootstrap-5-beta-3/) of bootstrap 5 has been released.
Thus, it's expected for the public beta to come in a few weeks.

So we can already start migrating on a different branch and get an early view of what breaks. 

## Further notes:
- See the [migration docs](https://getbootstrap.com/docs/5.0/migration/)
- jQuery support has been dropped, so we'll have to migrate a few of our js modules and test whether they still work.

lukasjuhrich:
Fixes #448 

Also includes a cleanup of the import compatability rules in the `webpack` config.

See also [the commits in wtforms-widgets](https://github.com/agdsn/wtforms-widgets/compare/94babca2...3d1ae00a)


lukasjuhrich:
Draft, because I forgot to go through the changelog for renamings I might've missed.

ibot3 (changes requested):
The changes look good. However I found two problems when testing on the branch:

- The level overview (e.g. `/facilities/building/7/level/1/rooms/`) isn't working for me. It seems that there is no AJAX request sent.
- The `account-typeahead` seems to be loaded on all pages which causes an unnecessary AJAX request to `/finance/json/accounts/system` on every page.

Idk if these problems are related to the changes of the PR, but at least the first one should be fixed (or shown that it is a problem on my side).

lukasjuhrich:
The `account-typeahead` request is not related to this PR, I have noticed it bebfore.
It prefetches the finance accounts, which works because there are less than 10 (contrary to multiple thousands of user accounts, which have to be fetched lazily, i.e. depending on user input).

lukasjuhrich:
I cannot reproduce the level overview issue. Please try to run `docker-compose run --rm dev-app webpack` and see whether this still happens after a hard reload (`C-S-r`)

lukasjuhrich:
Everything in the changelog already has been fixed. I did not observe other BS5-related issues.
Publishing && merging once CI works.

Issue #470, opened 2021-05-20
Identify users who could be archived


lukasjuhrich:
Child of #67.

## Purpose
We need a function that marks potential users as “to be archived”.
To test whether this function does the correct thing, and to make the planned process more transparent for the day-to-day users of pycroft, it would be helpful to make that filter visible by means of a quick UI table.

## Demands
- A function `get_users_to_be_archived()` in the `lib` (probably under `lib.user`, but in a sublayer of that, which would require splitting it up), and/or a function `should_be_archived()` asking that question about a single user
- A UI endpoint providing a table with all these users.

The following conditions should be met:
- the user has no room
- the user has no membership with the ~~`keep-alive`~~ `do-not-archive` flag now or in the future

lukasjuhrich:
What this task set out to do, namely defining a filter and providing an overview to judge that filter's effect, is done (#484).
The Celery task implementation && tagging theh users with `to-be-archived` shall happen in a folllow-up task.

Issue #52, opened 2016-10-06
Remove redundancy in table's `cols`


lukasjuhrich:
_(this is mainly a self-reminder and partially unordered thoughts.)_

The columns for a table are defined in a redundant manner – one in the html template, and once in the view function providing the json endpoint, since it provides a json dict with the columns as keys. At least one should not have to change colums in the html template when changing the json view, but in the view function and the json view. This logic doesn't belong in the template imho.

For instance, compare https://github.com/agdsn/pycroft/blob/develop/web/blueprints/user/__init__.py#L220 to https://github.com/agdsn/pycroft/blob/develop/web/templates/user/_user_show_hosts.html#L7
## EDIT:

See [this gist](https://gist.github.com/lukasjuhrich/28d976c44fd38bfc5ca407f27632c192).

---
## First ideas

One could solve it with providing the colums in an attribute given in the view function (for the page, not the table's json endpoint).  Then, it can be refactored out without complicated dependencies since both cols are in the same file.

One could spin this idea further and provide an object (most likely `namedtuple(…)()`) that provides all the necessary information and might carry the endpoint's name as well. Our template would reduce to

``` python
{% import "macros/table.html" as tbl %}
{{ tbl.table_from_object(host_table_info) }}
{# dir(host_table_info) contains {'columns': ['foo', 'bar'], 'json_endpoint': 'user.user_show_hosts_json'} #}
```

where `table_from_object` is a macro using `item_table` with the information from `host_table_info`.

the json view function might make use of some connected class like this:

``` python
@bp.route("/<int:user_id>/hosts")
def user_show_hosts_json(user_id):
    list_items = []
    for user_host in User.q.get(user_id).user_hosts:
        # …
        for ip in user_host.ips:
            # use the classmethod create_json_dict:
            list_items.append(UserHostsTableInfo.create_json_dict(
                id=str(user_host.id),
                ip=str(ip.address),
                mac=ip.interface.mac,
                switch=switches,
                port=ports,
                action=foo,
            ))
    return jsonify(items=list_items)
```

The redundancy is technically still there, but easier to test since it must match the function signature.

Perhaps this can be solved more elegantly still.

lukasjuhrich:
[...]

PR #484, opened 2021-07-24
Add query for archivable users and display them in table

lukasjuhrich:
This is a preliminary overview for which users would be marked as `to-be-archived`.

The next step would be to
1. Implement a lib function to mark these users as `to-be-archived` and filter out `to-be-archived` users in `get_archivable_members`
2. Split up this table into three tables with the same columns: 
  i. Users to-be-marked (i.e. what's returned from `get_archivable_members`)
  ii. Users who are `to-be-archived`
  iii. Users who are `to-be-archived` _and_ `do-not-archive` (i.e. invalidated memberships)
3. Implement a lib routine that actually archives the users, which fails if there are invalidated memberships

Refs #470

### DIFF
--- a/web/blueprints/user/__init__.py
+++ b/web/blueprints/user/__init__.py
@@ -1272,3 +1272,10 @@ def resend_confirmation_mail():
         session.session.commit()
 
         return redirect(return_url)
+
+
+@nav.navigate('Archivable users')
+@bp.route('/archivable_users')
+def archivable_users():
+    from pycroft.lib.user_deletion import get_archivable_members
+    return render_template('user/archivable_users.html', rows=get_archivable_members())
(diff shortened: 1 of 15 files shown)


web/blueprints/user/__init__.py
```diff
@@ -1272,3 +1273,33 @@ def resend_confirmation_mail():
...
+                title=info.User.room.short_name,
+                href=url_for('facilities.room_show', room_id=info.User.room.id)
+            ),
+            current_properties=" ".join(("~" if p.denied else "") + p.property_name
+                                        for p in info.User.current_properties_maybe_denied),
+            num_hosts=len(info.User.hosts),
+            # TODO better: `DateColumn.value`
+            end_of_membership=date_format(info.mem_end.date())
+        ) for info in get_archivable_members()
+    ]}
```
lukasjuhrich, line 1305: @ibot3 I would like feedback in particular about this strongly-typed way of constructing the table's columns. At least PyCharm was able to autocomplete the kwargs of `.value()` and `.row()`.
See #52 .

PR #491, opened 2021-08-04
Manual task exe

### LINKED ISSUE #465: Execute planned task immediately
Sometimes a task for relocation or move in exists but the member gets into the room early.
It would be nice to have a button to execute a task immediately instead of deleting the task and performing the action manually.

lukasjuhrich:
- Fixes #465
- Also adds a unit test for `cancel_task`

Screenshots: 
image
image

### DIFF
--- a/pycroft/lib/task.py
+++ b/pycroft/lib/task.py
@@ -5,6 +5,7 @@ from typing import Mapping, TypeVar, Generic
 from marshmallow import ValidationError
 from sqlalchemy.orm import with_polymorphic
 
+from pycroft.helpers.i18n import deferred_gettext
 from pycroft.lib.logging import log_task_event
 from pycroft.model import session
 from pycroft.model.session import with_transaction
@@ -210,4 +211,8 @@ def force_execute_task(task: Task, processor: User):
     if task.status != TaskStatus.OPEN:
         raise ValueError("Cannot execute a task that is not open")
 
-    pass  # not implemented
+    get_task_implementation(task).execute(task)
+
+    log_task_event(deferred_gettext("Manually executed task {}").format(task.id).to_json(),
+                   author=processor, task=task)
+    task.status = TaskStatus.EXECUTED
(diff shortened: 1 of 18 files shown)


pycroft/lib/task.py
```diff
@@ -192,10 +198,22 @@ def get_active_tasks_by_type(type):
...
     log_task_event("Cancelled task {}.".format(task.id), processor, task)
 
     task.status = TaskStatus.CANCELLED
+
+
+def manually_execute_task(task: Task, processor: User):
+    if task.status != TaskStatus.OPEN:
+        raise ValueError("Cannot execute a task that is not open")
+
+    get_task_implementation(task).execute(task)
```
ibot3, line 214: Catching errors here could be a good idea, because when an exception occurs, the user only sees a 500 error message.

ibot3 (approved):

pycroft/lib/task.py
lukasjuhrich: The question would be which errors to catch. A `except Exception:` would not be advisable. We want to have programming errors be fatal and reported.
However, if it's an error of the form „does not live in that room“ or sth, we will probably want to catch that.

But what does that mean to thhe lifecycle of the task? Does it go to `TaskState.FAILED`?

ibot3: I don't think that it is an good idea to reimplement the task execution logic here again. So we should either

- Do not change the state to `FAILED` but just display an error message instead

or

- Don't execute the task here directly, but instead trigger that the task should be executed by celery. For example by setting the due date to the current timestamp and calling `pycroft.task.execute_scheduled_tasks.delay()`.

lukasjuhrich: I think this is a good time to introduce a `PycroftException` for `lib` errors.

lukasjuhrich: See #493

lukasjuhrich: Resolved in f34fea5261b30ba55a6e5c5a5a00775a195427b3.

PR #492, opened 2021-08-04
Icons + Improved naming for user tabs

lukasjuhrich:
This changes some descriptions and turns the user tabs from this:
image
Into this:
image

I'd say this is a valuable change since I found myself scanning the page for the `Aufgaben` tab way too often. having a unified description and icons definitely helps me.
Also, I've noticed how my eyes first focused on the number badges, which provide the least important information, hence I've turned `bg-secondary` into `bg-light`.

### DIFF
--- a/web/templates/user/user_show.html
+++ b/web/templates/user/user_show.html
@@ -30,6 +30,7 @@
                {{ 'aria-disabled="true"' if 'disabled' in tab and tab.disabled else "" }}
                id="tab-{{ tab.id }}" href="#{{ tab.id }}"
                data-bs-toggle="tab" data-bs-target="#{{ tab.id }}">
+              {% if tab.icon %}<i class="fas {{ tab.icon }}"></i> {% endif -%}
               {{ tab.name }}
               {% if 'badge' in tab and tab.badge %}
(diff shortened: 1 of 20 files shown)


web/templates/user/user_show.html
```diff
@@ -30,9 +30,9 @@
                {{ 'aria-disabled="true"' if 'disabled' in tab and tab.disabled else "" }}
                id="tab-{{ tab.id }}" href="#{{ tab.id }}"
                data-bs-toggle="tab" data-bs-target="#{{ tab.id }}">
-              {{ tab.name }}
+              {{ tab.name | safe }}
```
ibot3, line 33: Adding a dedicated `icon` attribute to the tab object would be a better way in my opinion.

ibot3 (approved):

web/templates/user/user_show.html
lukasjuhrich: good point. That was the quick solution at the beginning where I was unsure how the result would look like.

lukasjuhrich:
Diff has been broken unfortunately because I rebased onto the other PR and pushed develop forward onto both branches at the same time, sorry about that.

PR #511, opened 2021-11-22
Resolve "Failed member verification (due to semicolon in last name)"

### LINKED ISSUE #510: Failed member verification (due to semicolon in last name)
In the rental contracts of Studentenwerk, some tenants have an additional semicolon concatenated to their last names.

In this case, our verification fails, as the prospective members only enter their real names (without the semicolon) to the registration form whereas the hashes are computed using the full names as specified in the rental contracts.

One possible counter measure would be, to try to add the semicolon to the last name, if the first try for verification fails, e.g.:
entered name: `Max Mustermann`
first try: `Max Mustermann`
second try: `Max Mustermann;`

LukasGibeh:
Closes #510

### DIFF
--- a/web/api/v0/__init__.py
+++ b/web/api/v0/__init__.py
@@ -449,6 +449,11 @@ class RegistrationResource(Resource):
 
         person_id = get_swdd_person_id(args.first_name, args.last_name, args.birthdate)
 
+        # some tenants have an additional semicolon added to their last names
+        if person_id is None:
+            person_id = get_swdd_person_id(
+                args.first_name, args.last_name + ";", args.birthdate)
+
         if person_id is None or person_id != args.person_id:
             abort(404, message="No tenancies found for this data",
                   code="no_tenancies")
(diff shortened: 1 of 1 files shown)


web/api/v0/__init__.py
```diff
@@ -449,6 +449,11 @@ def get(self):
 
         person_id = get_swdd_person_id(args.first_name, args.last_name, args.birthdate)
 
+        # some tenants have an additional semicolon added to their last names
+        if person_id is None:
+            person_id = get_swdd_person_id(
+                args.first_name, args.last_name + ";", args.birthdate)
```
lukasjuhrich, line 455:
```suggestion
            person_id = get_swdd_person_id(
                args.first_name, args.last_name + ";", args.birthdate
            )
```
If the opening bracket is followed by a line break, the closing bracket should be preceded by one. Other than that it's perfectly fine, thank you!

lukasjuhrich (approved):

web/api/v0/__init__.py
LukasGibeh: Hi @lukasjuhrich 

I applied the changes.

(Though, even the PEP8 standard is a bit unclear about this. You can find
```python
foo = long_function_name(
    var_one, var_two,
    var_three, var_four)
```
as well as 
```python
result = some_function_that_takes_arguments(
    'a', 'b', 'c',
    'd', 'e', 'f',
)
```
in there.)

Issue #136, opened 2017-12-21
Don't commit the session in view functions


lukasjuhrich:
There are several (>10) instances in `web.blueprints` where a manual commit of the session is issued. Such a manual intervention should not be necessary as this is the responsibility of the library `lib`.

sebschrader:
I don't necessarily agree. The HTTP request context/lifetime should also resemble the lifetime of the SQL transaction.

Abstracting transactions away from the user of the library API is not the right thing to do. This is not an implementation detail, it should rather be treated first class. The library doesn't know how many different updates the caller wants to perform in one transaction, only the user of the library does. The library might start it's own nested transactions.

In my opinion every endpoint that performs changes, should contain a with block:

```python
with session.begin():
   …
```

This gets even more important, if we enable higher isolation levels, in which transaction might fail due to locking issues. In this case, we might want to perform automatic retries.

One could certainly move most of it out into decorator and/or context manager, so that the simple endpoints don't have to implement his boilerplate on their own every time.

lukasjuhrich:
Let me phrase this differently.

There should currently be no **reason** for the view functions to wrap every library call with a `session.begin()`, because the library itself does not a priori know whether it is being called as a subroutine inside another library function (e.g. `log_user_event`), or resembles the top of the context stack. Most of the endpoints just resemble an interface to a particular library function such as `move_in`. IMO, it should not be necessary to set up a session just to make a lib call have an effect.
_however_, I totally agree with you that we probably should, independently of that, use a wrapping transaction anywhere to fit the request lifetime.

However, if something should actually break when removing such a `session.begin()`, this should be fixed.

sebschrader:
I don't agree. To quote [The Zen of Python](https://www.python.org/dev/peps/pep-0020/):

> Explicit is better than implicit.

I would actually prefer, that users of the library have to manage the session themselves, instead of all this implicit business. In this case, the lib functions could take the session as an argument, instead of importing a magic global session object. This would clearly indicate, who is responsible for managing the session. The user of the library must already set up this magic global session somehow beforehand, so she is responsible anyway, it's just hidden and implicit.

The library can then stop wrapping everything in transactions, except where it really wants to use  savepoints, because it's not its job anymore to manage transactions (it never should have been in the first place).

I know, this is what Flask does for its `request`, `session`, `app` etc. objects, but it has been criticized for exactly that (and IMHO rightly so). We won't and shouldn't change how Flask works, but for our code, we can decide, how we want to do things.

[...]

PR #519, opened 2021-12-01
Fix Editing membership: can't adapt type 'SimpleMessage' (#512)

go3ranh:
Issue #512

### DIFF
--- a/pycroft/lib/membership.py
+++ b/pycroft/lib/membership.py
@@ -10,13 +10,15 @@ management.
 
 """
 import typing as t
+from datetime import datetime
 
 from sqlalchemy import and_, func, distinct, Result
 from sqlalchemy.future import select
 from sqlalchemy.orm import aliased
 
+from pycroft.helpers import utc
 from pycroft.helpers.i18n import deferred_gettext
-from pycroft.helpers.interval import UnboundedInterval, IntervalSet, Interval
+from pycroft.helpers.interval import UnboundedInterval, IntervalSet, Interval, closedopen
 from pycroft.helpers.utc import DateTimeTz
 from pycroft.lib.logging import log_user_event, log_event
 from pycroft.model import session
@@ -203,3 +205,21 @@ def user_memberships_query(
         .group_by(Membership.id)
     )
     return session.session.execute(memberships)
+
+
+def change_membership_active_during(
+    membership_id: int, begins_at: datetime, ends_at: datetime | None, processor: User
+):
+    """
+    modify the active_during field of a user
+    """
+
+    membership: Membership = Membership.get(membership_id)
+    membership.active_during = closedopen(utc.with_min_time(begins_at), ends_at)
+
+    message = (
+        deferred_gettext("Edited the membership of group '{group}'. During: {during}")
+        .format(group=membership.group.name, during=membership.active_during)
+        .to_json()
+    )
+    log_user_event(message, processor, membership.user)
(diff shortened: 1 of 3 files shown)


pycroft/lib/membership.py
```diff
@@ -185,3 +189,25 @@ def user_memberships_query(user_id: int, active_groups_only: bool = False):
             .group_by(Membership.id)
     )
     return memberships
+
+
+def change_membership_active_during(
```
ibot3, line 194: Add `with_transaction` above.

pycroft/lib/membership.py
```diff
@@ -185,3 +189,25 @@ def user_memberships_query(user_id: int, active_groups_only: bool = False):
...
+        membership_id: int,
+        begins_at: datetime,
+        ends_at: Optional[datetime],
+        processor: User
+    ):
+    """
+    modify the active_during field of a user
```
ibot3, line 201: `... of a membership`

ibot3 (changes requested):

ibot3:
Make sure that the tests pass.

pycroft/lib/membership.py
lukasjuhrich: -1, we will deprecate this decorator at some point. Altering the session lifecycle should not be done in the lib. See [this discussion](https://github.com/agdsn/pycroft/issues/136#issuecomment-899034398). It makes things harder to control / reason about.
Perhaps I should add an ADR explaining the details.

lukasjuhrich:
Sorry for the logspam, I rebased this commit onto a more recent `develop` branch but forgot to push the latter beforehand.

lukasjuhrich:
@go3ranh if you would like to continue working on this, I've rebased this commit onto the current develop. If you have local commits, you can just type `git rebase bug_512 origin/bug_512` and that should do the trick.

You might notice that we now have substantially better CI feedback, especially the `lint` and `pre-commit` pipelines are helpful.

Also, I think we can simplify the test a lot by using `pytest`.

If you don't want to continue working on this just give me a heads up and someone else can complete the cleanup / expand on the test.

pycroft/lib/membership.py
lukasjuhrich: Actually this is sufficiently commented in [ADR004](https://agdsn.github.io/pycroft/arch/adr-004.html#cost-vs-benefit). The real thing missing here is to put `session` as the first parameter, but that wouldn't make any sense as of yet because `log_user_event` does not use the session explicitly.
So I guess one should fix it at another time, switching the paradigm for more than just one method.

pycroft/lib/membership.py
lukasjuhrich: Let's use fancy features to make it a little easier for @go3ranh :]
```suggestion
    modify the active_during field of a membership
 ```

File: pycroft/lib/membership.py [hunk 3]
```diff
@@ -200,3 +204,25 @@ def user_memberships_query(user_id: int, active_groups_only: bool = False) -> Qu
...
     )
     return memberships
+
+
+def change_membership_active_during(
+        membership_id: int,
+        begins_at: datetime,
+        ends_at: Optional[datetime],
+        processor: User
+    ):
```
lukasjuhrich, line 214: The „missing return type annotation“ mypy complains about means adding `-> None` to signify that nothing is returned.
Together with more [`black`-friendly styling](https://black.readthedocs.io/en/stable/the_black_code_style/current_style.html#how-black-wraps-lines) this would amount to
```suggestion
def change_membership_active_during(
    membership_id: int,
    begins_at: datetime,
    ends_at: Optional[datetime],
    processor: User,
) -> None:
```

Issue #493, opened 2021-08-06
Replace builtin exception raises by `PycroftException`s


lukasjuhrich:
Lots of our logic, be it in the `lib` or the `model`, uses exceptions to convey errors.

At some points, especially in older code, efforts have been started to use custom exception types such as `SubnetFullException` to convey meaning.

I started to introduce further structure by introducing a `PycroftException` → `PycroftLibException`/`PycroftModelException` hierarchy. However, when doing that I realized that the number of times we raise a builtin error is in the two-digit area.

Good examples are the exceptions defined in `lib/user.py` or the `CsvImportError` in `lib/finance.py` which even uses `gettext`.

**Pain:** 2
**Effort/Complexity:** 4
**Benefit:** 8
_(all on a scale from 1-10)_

lukasjuhrich:
Note that ideally, most errors would inherit from _both_ `Pycroft{Model,Lib}Exception` and a builtin exception such as `ValueError`, `TypeError`, `RuntimeError`, so that the rough category of error can still be communicated in a way that is not pycroft-specific.

PR #522, opened 2021-12-17
Proposal: Replace `web_execute` by context manager and base exception

lukasjuhrich:
@ibot3 Can you give me some feedback on this idea?
(Also relates to #493)

As mentioned previously, the value proposal is that calling functions directly can leverage IDE autocompletion and static analysis (i.e. type hints, or even just “knowing the function signature”). A call of the form `web_execute(func, success_message, *actual_arguments)` makes this impossible and is harder to reason about.

### DIFF
--- a/web/blueprints/user/__init__.py
+++ b/web/blueprints/user/__init__.py
@@ -648,18 +648,29 @@ def move(user_id):
     user = get_user_or_404(user_id)
     form = UserMoveForm()
 
-    refill_form_data = False
-    if form.validate_on_submit():
-        if user.room == Room.q.filter_by(
-                number=form.room_number.data,
-                level=form.level.data,
-                building_id=form.building.data.id).one():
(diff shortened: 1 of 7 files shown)


web/blueprints/user/__init__.py
```diff
@@ -538,59 +539,55 @@ def default_response():
...
+            )
+            sheet = lib.user.store_user_sheet(
+                True, wifi=(plain_wifi_password is not None),
+                user=new_user,
+                plain_user_password=plain_password,
+                plain_wifi_password=plain_wifi_password or ''
+            )
+            session.session.commit()
+    except PycroftException:
+        return default_response()
```
lukasjuhrich, line 585: Here is a good example for how the current proposal looks like

ibot3 (commented):
Looks like a great improvement of the current situation, I don't see anything to complain about :)

PR #534, opened 2022-01-31
set up semgrep

lukasjuhrich:
This PR
- Sets up [`semgrep`](https://semgrep.dev/docs/writing-rules/overview/) as part of `pre-commit`
- adds some sensible semgrep tests && fixes them:
  - migrations should not import production code (as it is subject to change)
  - `PasswordField`s should contain `password` or `secret` to help stupid sanitization regexes
  - log messages (e.g. those passed to `log_user_event`) should be formatted correctly using `deferred_gettext`

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -5,3 +5,9 @@ repos:
     hooks:
     -   id: pyupgrade
         args: ["--py310-plus"]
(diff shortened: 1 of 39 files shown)


lukasjuhrich:
CC @ibot3; perhaps you have some good ideas for static checks which we could add in the future.

PR #547, opened 2022-05-13
Ldap sync refactor

lukasjuhrich:
This PR more or less cleanly separates the mess that the ldap_exporter has grown into.

## Rationale
The current ldap syncer was hard to reason about due to a high amonunt of parameters in the outer methods and indirection.
Along the same lines, it proved fairly hard to add unit tests to the core logic.

## What happened
- extract a config datastructure
- and separate datastructures (Record, Action) from their behavior and conversion methods
- enable strict mypy checking for `ldap_sync`
- add unit tests for e.g. diff logic

It is best illustrated with the changes of module dependencies it introduces:

**Before**
image

**After**
image


lukasjuhrich:
Publishing it in spite of the legacy tests not being rewritten yet.
I'd rather have these changes upstream without that tackled than this becoming a WIP branch turning out-of-control, blocking development and bugfixes at the importer, and thus a time sink.

In this PR alone coverage by unit tests improved substantially.

Furthermore, the architecture should be substantially clearer.

ldap_sync/sources/db.py
```diff
@@ -0,0 +1,261 @@
...
+    [
+        "network_access",
+        "mail",
+        "traffic_limit_exceeded",
+        "payment_in_default",
+        "violation",
+        "member",
+        "ldap_login_enabled",
+        "userdb",
+        "cache_access",
```
ibot3, line 258: Can be removed

ibot3 (commented):
Logic not reviewed in detail, because it's too much.

But after a quick view, it looks good.

ldap_sync/sources/db.py
lukasjuhrich: Wait, why? This field is used in line 220, and logically, we have to restrict which properties we export

lukasjuhrich: Oh, you meant `cache_access`. Alright.
```suggestion
```

PR #554, opened 2022-05-27
Tempus dominus

### LINKED ISSUE #83: Use bootstrap-datetimepicker instead of datepicker
In some cases it is necessary to specify a datetime, not just a date.

lukasjuhrich:
Fixes #83 by configuring the [Tempus Dominus datetimepicker](https://getdatepicker.com/6/).

### DIFF
--- a/.eslintrc.yml
+++ b/.eslintrc.yml
@@ -1,10 +1,17 @@
 env:
   browser: true
   es2021: true
-extends: 'eslint:recommended'
+extends:
+  # TODO when everything is on ts, switch the rulesets
+  - "eslint:recommended"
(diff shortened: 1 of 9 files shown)


lukasjuhrich:
i rebased it, but i'm closing regardless: tempus-dominus is a) UX-wise a nightmare, and b) ISO8601-wise.
The latter can be fixed (although with a hacky plugin, not with a format), but not closing the picker after tabbing out of anything related is just not okay. I think it would be preferable at this point to look at [flatpickr](https://flatpickr.js.org/examples/) instead.

If anyone wants to fix these issues, the branch will still exist, but I don't have any plans to merge this PR.

PR #572, opened 2022-09-19
Fix error on cancellation of membership cancellation schedule

FestplattenSchnitzel:
Relates #512


lukasjuhrich:
@FestplattenSchnitzel thanks! In eb3670da4 I added type hints so mypy actually knows we need a `str` there, so this sort of bug cab be prevented in the future.

PR #598, opened 2023-01-28
Tombstoning

lukasjuhrich:
- [x] Implement UnixTombstone with FKeys, indices, and check constraints
- [x] Generic tests
- [x] Test check constraint
- [x] Test trigger creating Tombstone
- [x] lifetime restriction: Add `ON DELETE SET NULL` to `user→unix_account` (since the `unix_account` now does not serve as tombstone anymore) and ensure user deletion causes unix_account deletion
- [x] Think about whether it makes sense to keep `unix_accounts` without a `user` (the ldap_exporter ignores them, anyway) (**answer: automatically delete the unix account on user deletion, but technically allow them**)
- [ ] add migration for schema changes
- [x] proper schema documentation
- [x] vendor `mermaid.js`


pycroft/model/user.py
```diff
@@ -633,13 +648,247 @@ class Property(IntegerIdModel):
...
+          select ts.* into v_login_ts from "user" u
+              join unix_tombstone ts on v_u_login_hash = ts.login_hash
+              where u.id = NEW.id;
+
+          IF v_ua IS NULL THEN 
+              IF v_login_ts IS NULL THEN
+                  -- TODO check whether this was a _set_ or an _update_.
+                  -- do we really want to automatically update this?
+                  -- NOTE: when an update caused this, this might create an inconsistent state (different tombstones for uid and login),
+                  --  however as soon as the check constraint fires the transaction will be aborted, anyway.
```
lukasjuhrich, line 777: TODO: remove this comment
```suggestion
```

pycroft/model/user.py
```diff
@@ -633,13 +648,247 @@ class Property(IntegerIdModel):
...
+        language="plpgsql",
+    ),
+)
+
+manager.add_trigger(
+    User.__table__,
+    ddl.Trigger(
+        "user_ensure_tombstone_trigger",
+        User.__table__,
+        # TODO create different trigger on UPDATE which only fires if login or unix_account has changed
```
lukasjuhrich, line 802: TODO comment
```suggestion
```

lukasjuhrich:
The PR currently implements things like this:
```mermaid
erDiagram
    User {
        int unix_account_id
        str login
        str login_hash "generated always as digest(login, 'sha512')"
    } 
    UnixAccount {
        int id
        int uid
    }
    UnixTombStone {
        int uid
        str login_hash
    }

    User |o--o| UnixAccount: "…"
    User ||--o| UnixTombStone: "…"
    UnixAccount ||--o| UnixTombStone: "…"

```

lukasjuhrich:
Unfortunately, I cannot naturally create a cascade which deletes the `UnixAccount` if the user gets deleted – unless I flip the `User`←→`UnixAccount` dependency by adding a `UnixAccount.user_id` column. Any opinions on that @sebschrader @FestplattenSchnitzel?

lukasjuhrich:
> I cannot naturally create a cascade which deletes the `UnixAccount` if the user gets deleted – unless I […] [add] a `UnixAccount.user_id` column
Not true, I can just instruct sqlalchemy to issue the deletes by adding `cascade="delete"`.

tests/model/test_unix_tombstone.py
```diff
@@ -0,0 +1,246 @@
...
+        session.add(unix_account)
+        with pytest.raises(IntegrityError, match="ForeignKeyViolation"):
+            session.flush()
+
+
+class TestUserLoginHashFKey:
+    # TODO test user add, modify, delete
+    pass
+
+
```
lukasjuhrich, line 102:
```suggestion
 ```
Already covered in other tests.

ibot3 (approved):
I can't do an in-depth review currently but the changes look good for me.

lukasjuhrich:
Sorry for the abhorrent schema migration, but that's the way it is.

lukasjuhrich:
Ignoring the `pre-commit`, for some reason `darker` broke and could not merge. Probably because the PR commit length is above the clone depth, so `merge-base` failed.

PR #614, opened 2023-03-10
Fints errors

### LINKED ISSUE #529: Catch all `FintsError`s
Currently, FintsErrors other than `FinTSDialogError` and `FinTSClientPINError` result in a 500. We should instead catch the base exception:

https://github.com/raphaelm/python-fints/blob/846c25f5f6ae578cde5151fc30b5f7a3601a95ac/fints/exceptions.py#L5

Relevant section to extend:
https://github.com/agdsn/pycroft/blob/6da240db51183c3d0fccf616971bc90b3e8fff01/web/blueprints/finance/__init__.py#L202-L204

Low-hanging fruit, should be done before #527

lukasjuhrich:
Fixes #529 

- [x] rebase on top of `develop`
- [x] Make frontend view function more readable
- [x] Use `NamedTuple` for `process_transactions` view
- [x] Add tests for `pycroft.helpers.functional` (want 100% coverage)
- [x] Add tests for `pycroft.external_services.fints`


lukasjuhrich:
I think before merging it is advisable to give a name to the returned errors. In the fetching logic, there appears a variable `with_errors` whose meaning is elusive. Its type is `list[tuple[str, str]]`, and it is set here: https://github.com/agdsn/pycroft/blob/5538f91bfc19b32fb4246c1085a8bf2f325058fd/web/blueprints/helpers/fints.py#L54-L65

I prefer turning this into a `dataclass` so we know what the string entries of the typo mean. The variable name could be improved as well.

lukasjuhrich:
Also, I think for testability purposes this here should be extracted to a `lib` function returning statements and errors:
https://github.com/agdsn/pycroft/blob/5538f91bfc19b32fb4246c1085a8bf2f325058fd/web/blueprints/finance/__init__.py#L220-L238

Furthermore, we should look into why there needs to be so much manual form field setting going on. This looks like it should be defined on the form, not on the view function. I mean this:
https://github.com/agdsn/pycroft/blob/5538f91bfc19b32fb4246c1085a8bf2f325058fd/web/blueprints/finance/__init__.py#L197-L215
But also this:
https://github.com/agdsn/pycroft/blob/5538f91bfc19b32fb4246c1085a8bf2f325058fd/web/blueprints/finance/__init__.py#L183

It seems to me that there should an a) central and b) declarative way of stating such defaults.

PR #626, opened 2023-04-05
Update OpenSearch description

FestplattenSchnitzel:
- Add xmlns:moz namespace
- Add InputEncoding
- Add Image
- Fix typo

### DIFF
--- a/.mailmap
+++ b/.mailmap
@@ -26,3 +26,5 @@ Christian Jacobs <jacobs@wh2.tu-dresden.de>
 
 Tim Kluge <timklge@wh2.tu-dresden.de>
 <timklge@wh2.tu-dresden.de> <timklge@atlantis.wh2.tu-dresden.de>
+
+Gregor Düster <git@gdstr.eu> FestplattenSchnitzel <festplatte.schnitzel@posteo.de>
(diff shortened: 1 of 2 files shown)


mmarx:
> Add xmlns:moz namespace

why, though? No tags from the `moz` namespace are actually used in the description.

FestplattenSchnitzel:
> why, though?

Currently, it's not possible to add the Pycroft OpenSearch description.  Since almost all OSDs declare the xmlns:moz namespace, I guessed that might be the reason why it's not working currently.

mmarx:
It should only be required if you add the `moz:SearchForm` tag (shouldn't hurt to have it even if it's not needed, though). Another problem I can see is that the `Url` tag doesn't seem to allow a `method` attribute: https://github.com/dewitt/opensearch/blob/master/opensearch-1-1-draft-6.md#the-url-element

FestplattenSchnitzel:
`Url` with `method` attribute is used by Wikipedia view-source:https://de.wikipedia.org/w/opensearch_desc.php and works for me.

lukasjuhrich:
Perhaps you can test these hypotheses with a local nginx server to alleviate us from guesswork

FestplattenSchnitzel:
> Currently, it's not possible to add the Pycroft OpenSearch description.

OK, the reason for this was completely different: when adding an OSD, Firefox seems to make an isolated request unaware of any authentication (cookies, basic auth).
With the whole https://pycroft.agdsn.de/ deployment behind a basic auth, it ran into an HTTP 401 every time.

With https://git.agdsn.de/AGDSN/Ansible/-/merge_requests/195 , https://pycroft.agdsn.de/static/xml/opensearch.en.415ab08a4bb518e6a23d4fbf16e3c00b.xml has been excluded from basic auth, fixing the error.

Issue #596, opened 2023-01-11
Enable selection of multiple (not all) unconfirmed transactions for deletion / confirmation


FestplattenSchnitzel:
Here: https://pycroft.agdsn.de/finance/transactions/unconfirmed

lukasjuhrich:
This should not be too complicated. Bootstrap-table supports checkboxes: https://examples.bootstrap-table.com/index.html#methods/get-selections.html

The gist is:
1. You add a column like `<th data-field="state" data-checkbox="true"></th>`
2. You extract the selected rows with `getSelections`: `document.getElementById('table').bootstrapTable('getSelections')`
3. You issue a manual post request

Not sure if this is the most idiomatic way, but this is what I got from 5min of research.

PR #627, opened 2023-04-07
F/added checkbox for transactions

agmes4:
# Features 
selection for multiple transactions added 

# Outcome
added two columns one for id the other for select 
scheduling manual javascript post which posts the unconfirmed transaction ids to the backend 

closes: #596

### DIFF
--- a/web/blueprints/finance/__init__.py
+++ b/web/blueprints/finance/__init__.py
@@ -904,12 +904,12 @@ def transactions_confirm_selected():
     Confirms the unconfirmed transactions that where selected by the user in the frontend
     Javascript is used to post
     """
-    ids = request.form.getlist("ids[]")
+    ids = request.json.get("ids", [])
+
     for id in ids:
-        if id.isnumeric():
-            transaction = Transaction.get(int(id))
-            lib.finance.transaction_confirm(transaction, current_user)
-            session.commit()
+        transaction = Transaction.get(int(id))
+        lib.finance.transaction_confirm(transaction, current_user)
+        session.commit()
     return redirect(url_for('.transactions_unconfirmed'))
 
 
--- a/web/blueprints/finance/__init__.py
+++ b/web/blueprints/finance/__init__.py
@@ -897,15 +897,20 @@ def transaction_confirm(transaction_id):
     flash('Transaktion bestätigt.', 'success')
     return redirect(url_for('.transactions_unconfirmed'))
 
-@bp.route('/transaction/confirm_selected', methods=['GET', 'POST'])
+
+@bp.route('/transaction/confirm_selected', methods=['POST'])
 @access.require('finance_change')
 def transactions_confirm_selected():
-    ids = request.form.getlist("Ids", [])
-    print(ids)
+    """
+    Confirms the unconfirmed transactions that where selected by the user in the frontend
+    Javascript is used to post
+    """
+    ids = request.form.getlist("ids[]")
     for id in ids:
-        transaction = Transaction.q.filter_by(confirmed=False).filter(id=id)
-        transaction_confirm(transaction, current_user)
-        session.commit()
+        if id.isnumeric():
+            transaction = Transaction.get(int(id))
+            lib.finance.transaction_confirm(transaction, current_user)
+            session.commit()
     return redirect(url_for('.transactions_unconfirmed'))
 
 @bp.route('/transaction/confirm', methods=['GET', 'POST'])
--- a/web/blueprints/finance/__init__.py
+++ b/web/blueprints/finance/__init__.py
@@ -830,6 +830,7 @@ def transactions_unconfirmed_json():
 
         items.append(
             {
+                'id': transaction.id,
                 'description': T.description.value(
                     href=url_for(".transactions_show", transaction_id=transaction.id),
                     title=transaction.description,
@@ -896,6 +897,16 @@ def transaction_confirm(transaction_id):
     flash('Transaktion bestätigt.', 'success')
     return redirect(url_for('.transactions_unconfirmed'))
 
+@bp.route('/transaction/confirm_selected', methods=['GET', 'POST'])
+@access.require('finance_change')
+def transactions_confirm_selected():
+    ids = request.form.getlist("Ids", [])
+    print(ids)
+    for id in ids:
+        transaction = Transaction.q.filter_by(confirmed=False).filter(id=id)
+        transaction_confirm(transaction, current_user)
+        session.commit()
+    return redirect(url_for('.transactions_unconfirmed'))
 
 @bp.route('/transaction/confirm', methods=['GET', 'POST'])
 @access.require('finance_change')
--- a/web/blueprints/finance/__init__.py
+++ b/web/blueprints/finance/__init__.py
@@ -810,7 +810,6 @@ def transactions_unconfirmed():
         unconfirmed_transactions_table=UnconfirmedTransactionsTable(
             data_url=url_for(".transactions_unconfirmed_json"))
     )
-"""TODO: Add post mapping which forges the Payload send by JS and accepts the selected"""
 
 @bp.route('/transactions/unconfirmed/json')
 def transactions_unconfirmed_json():
--- a/web/blueprints/finance/__init__.py
+++ b/web/blueprints/finance/__init__.py
@@ -829,7 +829,7 @@ def transactions_unconfirmed_json():
 
         items.append(
             {
-                'id': transaction.id,
+"id": transaction.id,
                 'description': T.description.value(
                     href=url_for(".transactions_show", transaction_id=transaction.id),
                     title=transaction.description,
--- a/web/blueprints/finance/__init__.py
+++ b/web/blueprints/finance/__init__.py
@@ -904,7 +904,7 @@ def transactions_confirm_selected():
     Confirms the unconfirmed transactions that where selected by the user in the frontend
     Javascript is used to post
     """
-    ids = request.json.get("ids", [])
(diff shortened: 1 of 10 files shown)


web/blueprints/finance/__init__.py
```diff
@@ -810,7 +810,7 @@ def transactions_unconfirmed():
         unconfirmed_transactions_table=UnconfirmedTransactionsTable(
             data_url=url_for(".transactions_unconfirmed_json"))
     )
-
+"""TODO: Add post mapping which forges the Payload send by JS and accepts the selected"""
```
lukasjuhrich, line 813: `TODO` comment, please fix or remove (+track via issue if relevant)
```suggestion
```

Nota bene: Use comments for this (`# TODO`) instead of a multiline string literal (`"""TODO"""`), as there is no need to create a python string object.

However, for dostrings, it is correct to use a string instead of a comment, because strings which are specified directly below a function `f()` are assigned to its `f.__doc__` attribute, which makes `help` and autogeneration of API documentation (e.g. via `sphinx`) work.

web/blueprints/finance/__init__.py
```diff
@@ -830,6 +830,7 @@ def transactions_unconfirmed_json():
 
         items.append(
             {
+                'id': transaction.id,
```
lukasjuhrich, line 833: Not quite sure how this went through `darker`; the implicit convention for new code is double quotes instead of single quotes.
```suggestion
"id": transaction.id,
```

web/blueprints/finance/__init__.py
```diff
@@ -869,7 +870,8 @@ def transactions_unconfirmed_json():
                         icon='fa-trash',
                         btn_class='btn-danger btn-sm'
                     )
-                ] if privilege_check(current_user, 'finance_change') else [],
+                ]
+            if privilege_check(current_user, 'finance_change') else [],
```
lukasjuhrich, line 874: This is incorrect/misleading indentation, because the `if` part only removes the `actions` entry, and no other columns of the row. 
Before I suggest something that's incorrect in the eyes of `black`/`darker`, just throw `darker` on this file.

web/resources/js/transaction-form.js
```diff
@@ -63,4 +63,27 @@ $(function () {
...
+          success: function (result) {
+             console.log(result);
+          },
+          error:function (error){
+              console.log(error);
+          },
+          dataType: "json"
+        });
+      setTimeout(() => { window.location.reload(); }, 500);
+  })
```
lukasjuhrich, line 86: It would be nice if you could try to avoid using jquery and use the more modern (and, more importantly, builtin) [`fetch` API](https://developer.mozilla.org/en-US/docs/Web/API/fetch).

See e.g. this sipa code to get a feeling for how it works: https://github.com/agdsn/sipa/blob/cdd8d40a977c3ea2c70d491ec9e56b2d8935946d/sipa/static/js/statuspage.js#L47-L57

lukasjuhrich: this only concerns the AJAX part, I'm aware that bootstrap-table requires access via jquery

web/templates/finance/transactions_unconfirmed.html
```diff
@@ -9,4 +12,10 @@
         <span class="fa fa-check"></span>
         Alle bestätigen
     </a>
+  <button class="btn btn-success" id="accept_selected">Auswahl bestätigen</button>
 {% endblock %}
+
+
+  {% block page_script %}
+{{ resources.link_script_file('transaction-form.js' | require) }}
+    {% endblock %}
```
lukasjuhrich, line 21:
```suggestion
{% block page_script %}
{{ resources.link_script_file('transaction-form.js' | require) }}
{% endblock %}
```

web/blueprints/finance/__init__.py
```diff
@@ -896,6 +898,21 @@ def transaction_confirm(transaction_id):
     return redirect(url_for('.transactions_unconfirmed'))
 
 
+@bp.route('/transaction/confirm_selected', methods=['GET', 'POST'])
```
lukasjuhrich, line 901: Since you don't render a form here, there's no need for `GET`:
```suggestion
@bp.route("/transaction/confirm_selected", methods=["POST"])
```

lukasjuhrich:
Great job in general, seems like you figured out the most complicated parts!

web/resources/js/transaction-form.js
```diff
@@ -63,4 +63,27 @@ $(function () {
...
+          data: {ids: ids},
+          success: function (result) {
+             console.log(result);
+          },
+          error:function (error){
+              console.log(error);
+          },
+          dataType: "json"
+        });
+      setTimeout(() => { window.location.reload(); }, 500);
```
lukasjuhrich, line 85: What's the precise workflow here? Why do you need a timer of `0.5s`? I was thinking you select the transactions and then press something. Perhaps i misunderstood the whole workflow.

web/blueprints/finance/__init__.py
agmes4: the tested failed when I removed that get and then I was like it is how it is probably cause all the outer functions used get an redirect to select the specified transaction

web/resources/js/transaction-form.js
agmes4: when reloading the page instantly the old transactions will appear again even tough they are confirmed maybe cause the query is not fast enough and I should just follow the the redirects but that did not worked some how. I try this with the other post method again without that

lukasjuhrich: I see, let me take a look.

lukasjuhrich: A clean way would be to wait for the request to finish by issuing a callback. That's precisely what the `fetch()` API is very good at, because it returns promises you can handle with `.then()`:
```js
      if (ids.length > 0) {
          fetch("/finance/transaction/confirm_selected", {
              method: "POST",
              body: JSON.stringify({ids: ids}),
              headers: { "Content-Type": "application/json" },
          })
              .then(r => {
                  console.debug("Got response, reloading page");
                  window.location.reload();
              })
              .catch(e => console.error(`Got error when submitting transactions to confirm: ${e}`));
 ```
 Haven't tested it, but that should be the Idea.

File: web/blueprints/finance/__init__.py (hunk 6 above)
Comment (lukasjuhrich) [reply web/blueprints/finance/__init__.py:900]: Ah, I see, that's an oversight in the permissions test. It always uses `GET`… let me see whether I can fix that quickly.

File: web/resources/js/transaction-form.js (hunk 7 above)
Comment (agmes4) [reply web/resources/js/transaction-form.js:85]: when I used ajax I was not able to get a List when JSON.stringifying it i will test it

Comment (lukasjuhrich) [reply web/resources/js/transaction-form.js:85]: Ah, right, you need to pass it as `formdata` if you access it via `request.form`. You will need to construct a `FormData` object and pass it in the `body` parameter of the `fetch` function (see [MDN](https://developer.mozilla.org/en-US/docs/Web/API/fetch#body))

File: web/blueprints/finance/__init__.py (hunk 2 above)
Comment (lukasjuhrich) [reply web/blueprints/finance/__init__.py]: I haven't noticed that github trimmed the whitespace in the copy'n'paste process. Fixing this in a commit.
I now realized why `darker` did not catch this: the `pre-commit` hook only ran on the last commit of this PR instead of the whole range. I've fixed this in 999e1f02b79c4f26b782229e982b0da0c3baa127.

**Comment by lukasjuhrich** [comment]
I've finally fixed the `pre-commit` check in the CI. `pre-commit` is rather stupid and cannot produce markdown/github-annotations; However, you see the diff in the output:
```diff
--- web/blueprints/finance/__init__.py	2023-04-18 12:58:23.000000 +0000
+++ web/blueprints/finance/__init__.py	2023-04-18 16:35:21.640468 +0000
@@ -827,12 +827,12 @@
             transaction_id=transaction.id
         ).first()
 
         items.append(
             {
-"id": transaction.id,
-                'description': T.description.value(
+                "id": transaction.id,
+                "description": T.description.value(
                     href=url_for(".transactions_show", transaction_id=transaction.id),
                     title=transaction.description,
                     new_tab=True,
                     glyphicon='fa-external-link-alt'
                 ),
@@ -868,12 +868,14 @@
                                      transaction_id=transaction.id), title='Löschen',
                         icon='fa-trash',
                         btn_class='btn-danger btn-sm'
                     )
                 ]
-            if privilege_check(current_user, 'finance_change') else [],
-            })
+                if privilege_check(current_user, "finance_change")
+                else [],
+            }
+        )
 
     return jsonify(items=items)
 
 
 @bp.route('/transaction/<int:transaction_id>/confirm', methods=['GET', 'POST'])
@@ -891,12 +893,12 @@
 
     lib.finance.transaction_confirm(transaction, current_user)
 
     session.commit()
 
-    flash('Transaktion bestätigt.', 'success')
-    return redirect(url_for('.transactions_unconfirmed'))
+    flash("Transaktion bestätigt.", "success")
+    return redirect(url_for(".transactions_unconfirmed"))
 
 
 @bp.route("/transaction/confirm_selected", methods=["HEAD", "POST"])
 @access.require("finance_change")
 def transactions_confirm_selected():
```

lukasjuhrich:
Ideally you should run `pre-commit install` in the repo, try to fix the changes, and on commit the hook will tell you if you got it right. Fair warning: Initial installation can take a few minutes. But after that it's pretty fast.

lukasjuhrich:
I've fixed a typo and made some cosmetic changes in 481e3b38eeba43da3c5311a4fb7606e2eea69680, now this should be ready to merge.

PR #633, opened 2023-05-17
Fix logical error introduced in 165cc580bd68e6aef895019a5dc9b0764180d869

FestplattenSchnitzel:
A bank account activity is new, if *no* similar activity statement exists.


lukasjuhrich:
Next time please put the commit hash into the body, not the subject. Ignoring in the interest of time.

PR #635, opened 2023-05-25
B/fixed not vaild mac

agmes4:
# outcome 
Fixes the bug mentioned in #618
added some tests for the MAC address


lukasjuhrich (approved):
Everything looks great!
You got confused with the test parametrization, but after putting the index access on the right place everything passes immediately.

PR #640, opened 2023-06-17
Avoid HTML tags in plain text part of notification emails

FestplattenSchnitzel:


pycroft/templates/mail/user_reset_password.html
```diff
@@ -1,6 +1,6 @@
 {%- set links = [] -%}
 {%- import "macros/link.html" as link with context -%}
-<pre>
```
lukasjuhrich, line 3: If you import the macros as `link`, you probably need to use `link.start_pre`

lukasjuhrich:
Hey, so sorry for the late reply.

If I'm not mistaken, a more elegant way to conditionally attach `<pre>`s to every mail is to use `blocks` like we do with our `layout.html` (see the [documentation](https://jinja.palletsprojects.com/en/3.0.x/templates/#template-inheritance)):

```
{% extends "layout/mail.html" %}
{% block body %}
[…template content without <pre>]
{% endblock %}
```
And then in the `mail.html` you would use the `body` block and conditionally add the `<pre>`s depending on what `mode` is.

This is I think one line more, but it prevents you from doing stupid things like forgetting the `end_pre`.

FestplattenSchnitzel:
Any idea how I can test the changes?

ibot3:
> Any idea how I can test the changes?

Initially the `dev-mail` (mailslurper) docker compose service was meant to be used for mail testing.
It has a webinterface which is like an inbox.
However it sadly lacks some important features like viewing the source of the mail or switchting between html/plain.
(I am just seeing that it received the raw message feature with the [latest version](https://github.com/mailslurper/mailslurper/releases/tag/release-1.15.0))

An alternative is to add this docker compose service:

```
  mailserver:
    image: python:3.10
    command: python -u -m smtpd -n -c DebuggingServer 0.0.0.0:25
    stop_signal: SIGKILL
```

If configured as SMTP server in Pycroft, it prints all incoming mails to stdout of the container.

An other way that should probably be done either way is to write automated tests that check that the email template rendering works as expected.

FestplattenSchnitzel:
@ibot3 this doesn't seem to work with the current development setup and I doubt it's because of the `dummy-worker` (Celery) which doesn't do anything (?).  We should probably adjust the Docker composition to enable such tests …

FestplattenSchnitzel:
> If dev-celery-worker is running, sending mails should work.

Ah, thanks!

> An alternative is to add this docker compose service:

Unfortunately, the message parts (text, html) will still be base64 encoded.  Therefore, I wrote the script from the latest commit that prints the decoded versions.

Also, `dev-celery-worker` will print out all the mails (in encoded form) due to this line already: https://github.com/agdsn/pycroft/blob/d10b3c6642446363bdceb2b3aad954b769a707b1/pycroft/lib/mail.py#L86

lukasjuhrich:
Could we perhaps provide a cli command which renders the mail templates with certain flags set?

First, having a dedicated command that just renders the templates might be helpful to glance over it.

Second, just rendering the templates might be a good opportunity for a unit test, which checks a) if the template renders or crashes,  b) whether the HTML mode produces valid HTML, c) whether the plain text mode is valid plain text (think of a few heuristics; e.g. check whether the footnotes match the body). That way we could be pretty safe regarding a large class of issues.

docker/mail.Dockerfile
```diff
@@ -1,13 +1,8 @@
...
-RUN git clone "https://github.com/jamillosantos/mailslurper.git" /opt/mailslurper
-WORKDIR /opt/mailslurper/cmd/mailslurper
-RUN go get github.com/mjibson/esc
-RUN cd /opt/mailslurper/cmd/mailslurper
-COPY ./mailslurper.conf config.json
-RUN go get
-RUN go generate
-RUN go build
-ENTRYPOINT ["/opt/mailslurper/cmd/mailslurper/mailslurper"]
+FROM debian:unstable
```
ibot3, line 1: Maybe use `stable-slim` here?

lukasjuhrich:
@FestplattenSchnitzel in d9f1001cc0032070f17d4ed9b0c850ec45f016f4 I've added a fixture and assertion to the `move_in` tests. As we've discussed, just mocking the `user_send_mail[s]` seems to be a good option.

Ideally, something like this would exist for the other usages of `user_send_mail[s]`. A quick `ripgrep` run (at said commit hash) yields the following:
```
web/blueprints/user/__init__.py
1473:        lib.user.group_send_mail(

web/blueprints/finance/__init__.py
80:from pycroft.lib.user import encode_type2_user_id, user_send_mails
1538:            user_send_mails(negative_users, MemberNegativeBalance())

pycroft/lib/mail.py
95:def send_mails(mails: list[Mail]) -> tuple[bool, int]:
248:    from pycroft.task import send_mails_async
250:    send_mails_async.delay(mails)

pycroft/lib/user.py
67:from pycroft.task import send_mails_async
316:    user_send_mail(new_user, UserCreatedTemplate(), True)
408:    user_send_mail(user, UserMovedInTemplate(), True)
531:    user_send_mail(user, UserMovedInTemplate(), True)
1019:def user_send_mails(
1084:    send_mails_async.delay(mails)
1087:def user_send_mail(
1094:    user_send_mails([user], template, soft_fail, use_internal, **kwargs)
1106:def group_send_mail(group: PropertyGroup, subject: str, body_plain: str) -> None:
1108:    user_send_mails(users, soft_fail=True, body_plain=body_plain, subject=subject)
1114:    user_send_mail(
1132:    user_send_mail(user, UserConfirmEmailTemplate(
1345:            user_send_mail(mr, MemberRequestPendingTemplate(is_adult=mr.is_adult))
1383:        user_send_mail(prm, MemberRequestDeniedTemplate(reason=reason), soft_fail=True)
1500:        user_send_mail(user, UserResetPasswordTemplate(

pycroft/task.py
27:from pycroft.lib.mail import send_mails, Mail, RetryableException, \
124:    Errors are reported to the creator via `send_user_send_mail`.
186:    from pycroft.lib.user import user_send_mails
190:        user_send_mails(negative_users, MemberNegativeBalance())
196:        send_mails_async.delay([mail])
200:def send_mails_async(self, mails: list[Mail]):
205:        success, failures = send_mails(mails)
```

Also: I'm not sure if this is consistent, but the `test` part of the workflow now appears to be one minute faster… Really nice!

tests/lib/user/test_create_user.py
```diff
@@ -92,5 +93,16 @@ def test_finance_account(self, new_user):
...
+            "janed",
+            "jane.doe@example.org",
+            date.fromisoformat("2000-03-14"),
+            processor=processor,
+            groups=(member_group,),
+            address=room.address,
+        )
+
+        assert len(mail_capture) == 1
+        assert_mail_reasonable(mail_capture[0], subject_re="Willkommen")
```
lukasjuhrich, line 108:
```suggestion
        assert_mail_reasonable(assert_one(mail_capture), subject_re="Willkommen")
```

tests/lib/user/test_move.py
```diff
@@ -82,6 +86,9 @@ def test_moves_into_other_building(
         assert user.hosts[0].room == new_room_other_building
         # TODO test for changing ip
 
+        assert len(mail_capture) == 1
+        assert_mail_reasonable(mail_capture[0], subject_re="Wohnortänderung")
```
lukasjuhrich, line 90:
```suggestion
        assert_mail_reasonable(assert_one(mail_capture), subject_re="Wohnortänderung")
```

tests/lib/user/test_move.py
```diff
@@ -110,11 +117,14 @@ def full_params(self, new_room) -> dict[str]:
...
         }
 
-    def test_successful_move_execution(self, session, user, new_room, full_params):
+    def test_successful_move_execution(self, session, user, new_room, full_params, mail_capture):
         task = create_task_and_execute(TaskType.USER_MOVE, user, full_params)
         assert task.status == TaskStatus.EXECUTED
         assert user.room == new_room
 
+        assert len(mail_capture) == 1
+        assert_mail_reasonable(mail_capture[0], subject_re="Wohnortänderung")
```
lukasjuhrich, line 126:
```suggestion
        assert_mail_reasonable(assert_one(mail_capture), subject_re="Wohnortänderung")
```

lukasjuhrich: resolved in 6c1bde02bd47dcb4a1b5e09d541b9f2e7668c381

tests/lib/user/test_move.py
lukasjuhrich: resolved in 6c1bde02bd47dcb4a1b5e09d541b9f2e7668c381

tests/lib/user/test_create_user.py
lukasjuhrich: resolved in 6c1bde02bd47dcb4a1b5e09d541b9f2e7668c381

lukasjuhrich (approved):

PR #658, opened 2023-08-11
Improve view functions in `facilities` blueprint

lukasjuhrich:
- Upgrades to `flask~=2.3` (and `werkzeug>=2.0`)
- Improves the `handle_errors` context manager for more ergonomic usage (the `try` can be omitted, `session.begin_nested` needs to be called explicitly
- enforces stricter typing rules for the `facilities` blueprint
- Adds type hints to flask view functions across the board. This prevents stpuid bugs like forgetting to call `.model_dump` on a pydantic model.
- Enforces the latter via a semgrep rule.


web/app.py
```diff
@@ -146,11 +147,11 @@ def errorpage(e):
         return render_template('error.html', error=message), code
 
     @app.route('/')
-    def redirect_to_index():
+    def redirect_to_index() -> ResponseValue:
         return redirect(url_for('user.overview'))
 
     @app.route('/debug-sentry')
-    def debug_sentry():
+    def debug_sentry() -> ResponseValue:  # type: ignore
```
lukasjuhrich, line 154: I should be more specific with the suppression here.

web/blueprints/helpers/exception.py
```diff
@@ -39,12 +56,16 @@ def flash_and_wrap_errors() -> t.Iterator[None]:
         raise UnexpectedException from e
 
 
+ErrorHandlerMap = dict[type[PycroftException], ErrorHandler]
+
+
 @contextmanager
 # TODO rename to „wrap_errors“; `handle` suggests „I'll deal with everything“, which is incorrect
```
lukasjuhrich, line 63: I should think about a good name and ask people.

web/blueprints/facilities/__init__.py
```diff
@@ -503,18 +543,18 @@ def room_show(room_id):
...
 
 
 @bp.route('/room/<int:room_id>/patchpanel/json')
-def room_patchpanel_json(room_id):
-    room = Room.get(room_id)
+def room_patchpanel_json(room_id: int) -> ResponseReturnValue:
+    room = session.session.get(Room, room_id)
 
     if not room:
         abort(404)
```
lukasjuhrich, line 560: These `if room` checks should be extracted to a `get_room_or_404` method.

PR #661, opened 2023-08-17
Fully enable mypy for `web` package

lukasjuhrich:


pyproject.toml
```diff
@@ -56,6 +52,7 @@ module = [
     "web.blueprints",
     "web.blueprints.access",
     "web.blueprints.facilities",
+#    "web.blueprints.host",
```
lukasjuhrich, line 55: That should be reverted, but I am lazy and we want to look at enabling stricter rules for most modules, anyway

PR #684, opened 2024-01-21
fixed confirm justfile

agmes4:
the [confrim()] was not recognised there for I wrote a confirm clause which is run befor the Database was touched

### DIFF
--- a/justfile
+++ b/justfile
@@ -33,8 +33,15 @@ build:
     docker buildx bake
 
 # initializes the dev db with the instance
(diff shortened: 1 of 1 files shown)


lukasjuhrich:
Thanks! The confirm (especially with text) is a really, _really_ new feature of `just`. This is much more compatible.

I fixed a few details, see 4b8eb3c0b02bdc4b9b12ddc8420116a1435da5e4.

lukasjuhrich (approved):

Issue #631, opened 2023-04-27
Broken overcrowded view


gehinz:
The Page that shows overcrowded room just displays one of the accounts and isn't showing the second one anymore.

PR #685, opened 2024-01-21
fix overcrowded view see #631

jnsga:
to use itertools.groupby() the iterable needs to be sorted see https://docs.python.org/3/library/itertools.html#itertools.groupby


lukasjuhrich:
Amazing, thanks!

PR #686, opened 2024-01-21
changed justfile

agmes4:
added swdd.sql schema import for database

added for this also a create view

### DIFF
--- a/justfile
+++ b/justfile
@@ -34,9 +34,10 @@ build:
     docker buildx bake
 
 # initializes the dev db with the instance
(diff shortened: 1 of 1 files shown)


web/resources/js/user-suite.ts
lukasjuhrich, line 1: I guess this was committed by accident; I will remove this file in a rebase.

lukasjuhrich:
@agmes4 I adapted the recipe name to conform to the rest; in particular, I made it private, since you usually don't have to execute it yourself.

agmes4:
oh understand thanks :)

PR #688, opened 2024-01-25
just: Set file permission for .pycroft.pgpass, suggest depth 1 clone

FestplattenSchnitzel:


lukasjuhrich:
You can just push these changes to develop. I also noticed this just now.

lukasjuhrich:
I'll merge this after the python upgrade PR, which also fixes the pipeline (fixed and then broken in #687 because I used a py3.11 feature)

Issue #683, opened 2024-01-16
user overview page: make `→` select next tab


lukasjuhrich:
The most frequent navigational action on the user page is navigating to a different tab in the main tab pane (hosts, tasks, logs, …). It would be invaluable to bind the right arrow key (`→`) to „next tab“.

PR #691, opened 2024-01-26
added feature key navigation user suite

agmes4:
# Outcome 
added a key navigation on the user suite

solves #683

### DIFF
--- a/web/resources/js/user-suite.ts
+++ b/web/resources/js/user-suite.ts
@@ -1,0 +1,50 @@
+/*
+ * Copyright (c) 2024. The Pycroft Authors. See the AUTHORS file.
+ * This file is part of the Pycroft project and licensed under the terms of
+ * the Apache License, Version 2.0. See the LICENSE file for details
+ */
+
+function waitForKeyPress(callback: (event: KeyboardEvent) => void): void {
+  const handleKeyPress = (event: KeyboardEvent) => {
+    // Call the callback function with the event when a key is pressed
+    callback(event);
+
+    // Remove the event listener after the first key press (if you want it to happen only once)
+    // document.removeEventListener('keydown', handleKeyPress);
+  };
+
+  // Add the event listener to the document
+  document.addEventListener('keydown', handleKeyPress);
+}
+
+// Example of using the function
+waitForKeyPress((event) => {
+  console.log(`Key pressed: ${event.key}`);
+  // Add your logic here based on the key press
+    if (event.key === "ArrowRight" ){
+        const navLinks = document.getElementsByClassName("user-nav-link");
+        for (let i= 0; i < navLinks.length; i++){
+            const link = navLinks[i] as HTMLLinkElement;
+            if(link.classList.contains("active")) {
+                let index = (i + 1)%navLinks.length;
+                const link2 = navLinks[index] as HTMLLinkElement;
+                link2.click();
+                break;
+            }
+        }
+    }
+    if (event.key === "ArrowLeft" ){
+        const navLinks = document.getElementsByClassName("user-nav-link");
+        let next = false;
+        console.log(navLinks.length)
+        for (let i= 0; i < navLinks.length; i++){
+            const link = navLinks[i] as HTMLLinkElement;
+            if(link.classList.contains("active")) {
+                let index = (i - 1) % navLinks.length;
+                const link2 = navLinks[index] as HTMLLinkElement;
+                link2.click();
+                break;
+            }
+        }
+    }
+});
--- a/web/resources/js/user-suite.ts
+++ b/web/resources/js/user-suite.ts
@@ -22,7 +22,7 @@ waitForKeyPress((event) => {
   console.log(`Key pressed: ${event.key}`);
   // Add your logic here based on the key press
     if (event.key === "ArrowRight" ){
-        const navLinks = document.getElementsByClassName("user-nav-link");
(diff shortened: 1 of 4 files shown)


web/resources/js/user-suite.ts
```diff
@@ -0,0 +1,50 @@
...
+// Example of using the function
+waitForKeyPress((event) => {
+  console.log(`Key pressed: ${event.key}`);
+  // Add your logic here based on the key press
+    if (event.key === "ArrowRight" ){
+        const navLinks = document.getElementsByClassName("user-nav-link");
+        for (let i= 0; i < navLinks.length; i++){
+            const link = navLinks[i] as HTMLLinkElement;
+            if(link.classList.contains("active")) {
+                let index = (i + 1)% (navLinks.length - 1);
```
lukasjuhrich, line 29: shouldn't this be `(i + 1) % navLinks.length`? The indices lie in `{0, …, navLinks.Length - 1}`, right?

web/resources/js/user-suite.ts
```diff
@@ -0,0 +1,50 @@
...
+
+  // Add the event listener to the document
+  document.addEventListener('keydown', handleKeyPress);
+}
+
+// Example of using the function
+waitForKeyPress((event) => {
+  console.log(`Key pressed: ${event.key}`);
+  // Add your logic here based on the key press
+    if (event.key === "ArrowRight" ){
```
lukasjuhrich, line 24: Currently, this tab selection works even if you use ←/→ in an input field (e.g. a search field). to avoid this, add a pre-check like that:
```suggestion
    if (event.target.matches("input, textarea")) { return; }
    
    if (event.key === "ArrowRight" ){    
```

lukasjuhrich (commented):
The feature is really neat!
We can polish the JS a little here and there, and handle some edge-cases that I found.

web/resources/js/user-suite.ts
lukasjuhrich: But please check if it works, I'm not 100% sure whether the CSS selector on [`matches()`](https://developer.mozilla.org/en-US/docs/Web/API/Element/matches) works like that.

lukasjuhrich:
In a branch [`user_tabs_no_jumping`](https://github.com/agdsn/pycroft/tree/user_tabs_no_jumping) (commit 1355376c7afabb4a973e50d5434e432a716ac512), I've fixed the weird jumping that occurs when switching e.g. to/from the log tab. Much smoother I think.

You can cherry-pick this commit: `git remote update; git cherry-pick 1355376c7afabb4a973e50d5434e432a716ac512` to include it in this PR if you want.

web/resources/js/user-suite.ts
```diff
@@ -0,0 +1,50 @@
...
+        const navLinks = document.getElementsByClassName("user-nav-link");
```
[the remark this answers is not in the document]
lukasjuhrich: Ah, even cooler:
```suggestion
        const navLinks = [...document.querySelectorAll<HTMLElement>(".user-nav-link")].filter(x => !x.classList.contains("disabled)
```
Then you can remove the `as HTMLElement` cast below!

web/resources/js/user-suite.ts
```diff
@@ -0,0 +1,50 @@
...
+        }
+    }
+
+    if (event.key === "ArrowLeft" ){
+        const navLinks = document.getElementsByClassName("user-nav-link");
+        for (let i= 0; i < navLinks.length; i++){
+            const link = navLinks[i] as HTMLLinkElement;
+            if(link.classList.contains("active")) {
+                let index = i - 1;
+                if (index < 0) index += (navLinks.length - 1); // Ich hasse JS noch nicht mal mod kann das ordentlich
```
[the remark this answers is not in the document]
agmes4: when I get -1 % 8 = -1

web/resources/js/user-suite.ts
agmes4: it was not possible so I checked for isinstance

agmes4: But your totally right I have not thought about text areas

web/resources/js/user-suite.ts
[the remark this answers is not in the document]
lukasjuhrich: in that case, providing a reasonable `mod`  function in the module scope would be the way to go:
```js
const mod = (k, n) => ((k % n) + n) % n
```
And then just do `const index = mod(i-1, navLinks.length)`

web/resources/js/user-suite.ts
lukasjuhrich: see below, use a `mod` function that actually does modulo

web/resources/js/user-suite.ts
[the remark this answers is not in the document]
lukasjuhrich: Disregard that, there's no need for that nitpicking. resolving :)

web/resources/js/user-suite.ts
lukasjuhrich: Disregard that, there's no need for that nitpicking. resolving :)

lukasjuhrich:
@agmes4 if you're done with the changes, ping me, I'll test it briefly again and then I think we can merge :)

lukasjuhrich (approved):

agmes4:
I am done @lukasjuhrich  :)

PR #697, opened 2024-02-14
Fix default date for transaction creation

FestplattenSchnitzel:
Until now, the default date was the one the app was started on as the value was calculated on class initialisation.


lukasjuhrich:
Good catch! Run `darker` and it's fine to merge.

lukasjuhrich (approved):

PR #698, opened 2024-02-16
Add tenancies to room view

lukasjuhrich:
@hw0lff creating this PR for you so I can already leave a quick review before tomorrow.

### DIFF
--- a/web/blueprints/facilities/__init__.py
+++ b/web/blueprints/facilities/__init__.py
@@ -58,7 +58,7 @@ from web.blueprints.facilities.forms import (
 from web.blueprints.helpers.log import format_room_log_entry
 from web.blueprints.helpers.user import user_button
 from web.blueprints.navigation import BlueprintNavigation
-from web.table.table import TableResponse, LinkColResponse, BtnColResponse
+from web.table.table import TableResponse, LinkColResponse, BtnColResponse, date_format
 from .address import get_address_entity, address_entity_search_query
 from .tables import (
     BuildingLevelRoomTable,
@@ -70,9 +70,12 @@ from .tables import (
     BuildingLevelRoomRow,
     PatchPortRow,
     RoomOvercrowdedRow,
+    RoomTenanciesTable,
+    RoomTenanciesRow,
 )
 from ..helpers.exception import abort_on_error, ErrorHandlerMap
 from ..helpers.log_tables import LogTableRow
+from ...template_filters import date_filter
 
 bp = Blueprint('facilities', __name__)
 access = BlueprintAccess(bp, required_properties=['facilities_show'])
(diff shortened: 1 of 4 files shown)


web/blueprints/facilities/tables.py
```diff
@@ -104,3 +105,22 @@ class PatchPortRow(BaseModel):
...
+class RoomTenanciesTable(BootstrapTable):
+    _render_toolbar = False
+
+    inhabitant = BtnColumn("Bewohner")
+    begins_at = DateColumn("Von")
+    ends_at = DateColumn("Bis")
+    status = Column("Status")
+
+    class Meta:
+        table_args = {"data-sort-name": "begins_at"}
```
lukasjuhrich, line 119:
```suggestion
        table_args = {
            "data-sort-name": "begins_at",
            "data-search": "false",
        }
```

web/blueprints/facilities/__init__.py
```diff
@@ -519,6 +522,11 @@ def room_show(room_id: int) -> ResponseReturnValue:
     patch_port_table = PatchPortTable(data_url=url_for(".room_patchpanel_json", room_id=room.id),
                                       room_id=room_id)
 
+    room_tenancies_table = RoomTenanciesTable(
+        data_url=url_for(".room_tenancies_json", room_id=room.id),
+        table_args={"data-search": "false"},
```
lukasjuhrich, line 527: Move these table args to the `Meta` args of the table definition
```suggestion
```

web/blueprints/facilities/tables.py
```diff
@@ -104,3 +105,22 @@ class PatchPortRow(BaseModel):
...
+
+
+class RoomTenanciesRow(BaseModel):
+    inhabitant: BtnColResponse
```
lukasjuhrich, line 123: Why not `BtncolResponse | None` and allow showing tenancies with no users? Seems like most of the time this should not happen but if it does we would want to see it.

lukasjuhrich: @FestplattenSchnitzel any reason you resolved this and all the other comments of mine? were they not useful, or did this happen on accident?

lukasjuhrich: BTW a use case for this would be: there is a tenancy with a user (i.e. `person_id`), but the person id is not mapped to any user in our DB, e.g. because they did not register with us yet.

FestplattenSchnitzel: @lukasjuhrich I think I resolved the suggestions b/c they were implemented back then (https://github.com/agdsn/pycroft/commit/ede44d9578ad63d5c7430f373cfec4204679dc1c), but looks like you overrode the changes?

PR #701, opened 2024-02-27
Handle reversals for new imported transactions

### LINKED ISSUE #700: Handle credit reversals in bank account activities

FestplattenSchnitzel:
Closes #700

### DIFF
--- a/pycroft/lib/finance/transaction_crud.py
+++ b/pycroft/lib/finance/transaction_crud.py
@@ -221,9 +221,15 @@ def process_transactions(
         ):
             purpose = purpose + " EREF+" + transaction.data["end_to_end_reference"]
 
+        amount = transaction.data["amount"].amount
+
+        # Reversal ("Storno")
+        if "R" in transaction.data.get("status", ""):
+            amount = -amount
+
         new_activity = BankAccountActivity(
             bank_account_id=bank_account.id,
-            amount=transaction.data["amount"].amount,
+            amount=amount,
             reference=purpose,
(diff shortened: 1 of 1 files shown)


FestplattenSchnitzel:
Setting to draft b/c we should definitely add a test.

pycroft/lib/finance/transaction_crud.py
```diff
@@ -221,9 +221,15 @@ def process_transactions(
         ):
             purpose = purpose + " EREF+" + transaction.data["end_to_end_reference"]
 
+        amount = transaction.data["amount"].amount
+
+        # Reversal ("Storno")
+        if "R" in transaction.data.get("status", ""):
+            amount = -amount
+
```
lukasjuhrich, line 229: Instead of doing that inline, we should provide an intermediate function which turns a `MT940Transaction` into a `BankAccountActivity`. For that, it needs from `transaction.data` the following fields:
- `applicant_iban`
- `applicant_bin`
- `applicant_name`
- `purpose`
- `end_to_end_reference`
- `amount`
- `guessed_entry_date`
- `date`

This should be a pure function
```python
def activiity_from_transaction_data(
    transaction_data: TDataTypedDict,
    bank_account_id: int
) -> BankAccountActivity: ...
```
where `TDataTypedDict` captures the aforementioned keys as a `TypedDict`.
We can then use a fixed mt940 dump of test data to check whether all the cases are handled properly without even having to touch the DB.

PR #702, opened 2024-03-15
Enable export of non-attributable transfers

FestplattenSchnitzel:


pycroft/model/alembic/versions/bc0e0dd480d4_add_bankaccount_owner.py
```diff
@@ -0,0 +1,84 @@
...
+    op.drop_constraint("room_history_entry_user_id_fkey", "room_history_entry", type_="foreignkey")
+    op.create_foreign_key(
+        None, "room_history_entry", "room", ["room_id"], ["id"], ondelete="CASCADE"
+    )
+    op.create_foreign_key(
+        None, "room_history_entry", "user", ["user_id"], ["id"], ondelete="CASCADE"
+    )
+    op.alter_column("task", "task_type", existing_type=sa.VARCHAR(length=50), nullable=False)
+    op.alter_column("traffic_volume", "user_id", existing_type=sa.INTEGER(), nullable=False)
+    # ### end Alembic commands ###
```
lukasjuhrich, line 51: The only statement which should be there is the first one (adding `owner`). Do we have such drastic divergences between theoretical `model` and the actual model in the DB?

FestplattenSchnitzel: > The only statement which should be there is the first one (adding owner).

Removed the other ones.

> Do we have such drastic divergences between theoretical model and the actual model in the DB?

I don't know.  I don't know how the generation of this worked exactly either.

web/blueprints/finance/__init__.py
```diff
@@ -548,6 +552,66 @@ def bank_account_activities_match() -> ResponseReturnValue:
...
+    form: t.Any = _create_form(field_list)
+
+    return render_template(
+        "finance/bank_account_activities_return.html",
+        form=form(),
+        activities=activities,
+    )
+
+
+@bp.route("/bank-account-activities/return/do/", methods=["GET", "POST"])
```
FestplattenSchnitzel, line 589: The `"GET"` is just there so the `tests/frontend/test_permissions.py:test_access_finance` receives a `403` instead of a `405`.  Do you think that's fine?

FestplattenSchnitzel: (or should the test be adjusted?)

pycroft/helpers/printing/__init__.py
```diff
@@ -289,7 +290,7 @@ def generate_user_sheet(
             six monthly contributions at once.'''.format(
             contribution / 100), style['JustifyText']))
 
-        recipient = 'Studierendenrat TUD - AG DSN'
+        recipient = bank_account.owner
```
lukasjuhrich, line 293: Perhaps it would be wise to emit a warning in case `owner` is `None ` so we see it in sentry. Sure, it shouldn't happen in practice, but the types don't quite match, right? So in particular perhaps set recipient to the empty string in that case so we can be sure the rendering doesn't break.

web/blueprints/finance/__init__.py
lukasjuhrich: The test, whose implicit assumption is that every endpoint is `GET`, should definitely be adjusted, as this assumption is wrong.

web/templates/finance/bank_account_activities_return.html
```diff
@@ -0,0 +1,46 @@
+{#
+SPDX-FileCopyrightText: 2024 Gregor Düster <git@gdstr.eu>
+
+SPDX-License-Identifier: Apache-2.0
+#}
```
lukasjuhrich, line 5: Can you remove this header and/or replcace it with the header we usually use?
I am open to using more formal headers but that is a separate discussion.
(Also perhaps enforcing this via pre-commit hook might be a good idea)

pycroft/lib/finance/retransfer.py
```diff
@@ -0,0 +1,47 @@
...
+    sepa = SepaTransfer(transfer_config, clean=False)
+
+    for activity in activities:
+        payment = {
+            "name": activity.other_name,
+            "IBAN": activity.other_account_number,
+            "BIC": activity.other_routing_number,
+            "amount": int(activity.amount * 100),
+            "execution_date": datetime.now().date(),
+            "description": f"Rücküberweisung nicht zuordenbarer Überweisung vom {activity.posted_on} mit Referenz {activity.reference}",
```
lukasjuhrich, line 43: Wieder was [gelernt](https://www.duden.de/rechtschreibung/zuordenbar)!

pycroft/lib/finance/retransfer.py
```diff
@@ -0,0 +1,47 @@
...
+        .options(joinedload(BankAccountActivity.bank_account))
+        .filter(BankAccountActivity.transaction_id.is_(None))
+        .filter(BankAccountActivity.amount > 0)
+        .filter(BankAccountActivity.imported_at < ensure_tz(datetime.utcnow() - timedelta(days=14)))
+    )
+
+    return session.session.scalars(statement).all()
+
+
+def generate_activities_return_sepaxml(activities: list[BankAccountActivity]) -> bytes:
```
lukasjuhrich, line 26: This method needs a test – which is easy because you separated concerns well: this function has no implicit state nor side effects.

In this case I would actually just save the whole XML response in a test file, verify it visually, and access it via [`pkg_resources`](https://setuptools.pypa.io/en/latest/pkg_resources.html#basic-resource-access).

pycroft/lib/finance/retransfer.py
```diff
@@ -0,0 +1,47 @@
...
+
+
+def get_activities_to_return() -> Sequence[BankAccountActivity]:
+    statement = (
+        select(BankAccountActivity)
+        .options(joinedload(BankAccountActivity.bank_account))
+        .filter(BankAccountActivity.transaction_id.is_(None))
+        .filter(BankAccountActivity.amount > 0)
+        .filter(BankAccountActivity.imported_at < ensure_tz(datetime.utcnow() - timedelta(days=14)))
+    )
```
lukasjuhrich, line 21: Here some simple test cases (one for each filter) would be nice as well.

pycroft/lib/finance/retransfer.py
```diff
@@ -0,0 +1,47 @@
...
+
+    return session.session.scalars(statement).all()
```
lukasjuhrich, line 23: Also, please explicitly pass the `session` as per [ADR004](https://agdsn.github.io/pycroft/arch/adr-004.html):
```suggestion
def get_activities_to_return(session: Session) -> Sequence[BankAccountActivity]:
    statement = (
        select(BankAccountActivity)
        .options(joinedload(BankAccountActivity.bank_account))
        .filter(BankAccountActivity.transaction_id.is_(None))
        .filter(BankAccountActivity.amount > 0)
        .filter(BankAccountActivity.imported_at < ensure_tz(datetime.utcnow() - timedelta(days=14)))
    )

    return session.scalars(statement).all()
```

pycroft/lib/finance/retransfer.py
FestplattenSchnitzel: Also see the [entry in the DWDS](https://www.dwds.de/wb/zuordenbar) for those who like OpenSearch and dislike cookie banners, tracking and third party JS. ;)

pycroft/helpers/printing/__init__.py
FestplattenSchnitzel: I've set the server_default in the migration to `""` now.  This won't work for exporting non-attributable transfers though, some longer name has to be set.

lukasjuhrich (approved):

PR #708, opened 2024-04-02
Room tenancies, find lost commits

hw0lff:
With this PR, the person_id is also shown in the webview, even if no user could be mapped.

### DIFF
--- a/web/blueprints/facilities/__init__.py
+++ b/web/blueprints/facilities/__init__.py
@@ -617,6 +617,7 @@ def room_tenancies_json(room_id: int) -> ResponseReturnValue:
         items=[
             RoomTenanciesRow(
                 inhabitant=user_button(tenancy.user) if tenancy.user else None,
+                swdd_person_id=tenancy.person_id if tenancy.person_id else None,
                 begins_at=date_format(tenancy.mietbeginn, formatter=date_filter),
                 ends_at=date_format(tenancy.mietende, formatter=date_filter),
                 status=tenancy.status.name,
--- a/web/blueprints/facilities/__init__.py
+++ b/web/blueprints/facilities/__init__.py
@@ -533,7 +533,7 @@ def room_show(room_id: int) -> ResponseReturnValue:
         page_title=f"Raum {room.short_name}",
         room=room,
         ports=room.patch_ports,
-        user_buttons=[user_button(user).model_dump() for user in room.users],
+        user_buttons=[user_button(user) for user in room.users],
         user_histories=[
             (
                 user_button(room_history_entry.user).model_dump(),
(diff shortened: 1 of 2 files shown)


web/blueprints/facilities/__init__.py
```diff
@@ -613,10 +613,18 @@ def room_patchpanel_json(room_id: int) -> ResponseReturnValue:
 @bp.route("/room/<int:room_id>/tenancies/json")
 def room_tenancies_json(room_id: int) -> ResponseReturnValue:
     room = get_room_or_404(room_id)
+
+    def tenancy_person_id(tenancy) -> int | None:
+        return tenancy.person_id if tenancy.person_id is not None else None
+
+    def tenancy_user_button(tenancy) -> BtnColResponse | None:
+        return user_button(tenancy.user) if tenancy.user is not None else None
+
```
lukasjuhrich, line 622: Using nested functions with _one_ invocation is nonsense unless you have a really good reason. Here, you could've used a plain variable if you don't like the line length.

Otherwise, if we want to do null coalescence anyway, apart from fancy solutions we could just bake the null coalescence into `user_button` (changing its signature to `User | None -> str | None`)

hw0lff: I'm not sure a variable would be applicable here since it is used in a list comprehension. Avoiding long lines was the intention but let's inline these functions if there is not a better way to do it.

lukasjuhrich: Right, sorry, I overlooked the list comprehension. Then I'm not sure and I would just vouch for inlining, purely because weakening `user_button` changes semantics.

lukasjuhrich: I mean the other way would be the fancy bs way:
```python
def lift[A, B](f: Callable[[A], B | None]) -> Callable[[A | None], B | None]:
    def _lifted(a: A | None) -> B | None:
        return f(a) if a is not None else None
    return _lifted
```
and then use `lift(user_button)(tenancy.user)`

lukasjuhrich: (beware, untested)

web/blueprints/facilities/__init__.py
```diff
@@ -613,10 +613,12 @@ def room_patchpanel_json(room_id: int) -> ResponseReturnValue:
 @bp.route("/room/<int:room_id>/tenancies/json")
 def room_tenancies_json(room_id: int) -> ResponseReturnValue:
     room = get_room_or_404(room_id)
+
     return TableResponse[RoomTenanciesRow](
         items=[
             RoomTenanciesRow(
-                inhabitant=user_button(tenancy.user) if tenancy.user else None,
+                inhabitant=user_button(tenancy.user) if tenancy.user is not None else None,
+                swdd_person_id=tenancy.person_id if tenancy.person_id is not None else None,
```
hw0lff, line 621: Can the condition `if tenancy.user is not None else None` be shortened to `if tenancy.user else None` or does it change the meaning?
Same question for `tenancy.person_id` in line 621.

lukasjuhrich: Yes, since sqlalchemy objects are falsy iff they are `None`.

PR #725, opened 2024-05-18
Always enable splitting during bundling

FestplattenSchnitzel:
Apparently issues appear when bundling Bootstrap in multiple files.

### DIFF
--- a/web/resources/js/tab-anchor.ts
+++ b/web/resources/js/tab-anchor.ts
@@ -26,7 +26,7 @@ function updateLocationHash(this: HTMLElement, _: MouseEvent){
     }
 }
 
-document.addEventListener('DOMContentLoaded', () => {
+window.addEventListener('load', () => {
     navigateToAnchorTab();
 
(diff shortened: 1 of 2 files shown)


FestplattenSchnitzel:
Please verify that this doesn't just work for me.

web/resources/js/tab-anchor.ts
```diff
@@ -26,7 +26,7 @@ function updateLocationHash(this: HTMLElement, _: MouseEvent){
     }
 }
 
-document.addEventListener('DOMContentLoaded', () => {
+window.addEventListener('load', () => {
     navigateToAnchorTab();
```
FestplattenSchnitzel, line 30: This line works at `DOMContentLoaded` already, we could add another event listener if desired.

lukasjuhrich:
> Apparently issues appear when bundling Bootstrap in multiple files.

you mean _disappear_?
Not happy with this result as it isn't really an answer, but if it works it works. Will verify in a sec.

PR #739, opened 2024-07-08
Lookup missing BIC when returning bank account activities

FestplattenSchnitzel:


pycroft/lib/finance/retransfer.py
```diff
@@ -46,3 +51,27 @@ def generate_activities_return_sepaxml(activities: list[BankAccountActivity]) ->
...
+            author=author,
+            valid_on=activity.valid_on,
+            confirmed=False,
+        )
+        activity.split = next(
+            split for split in transaction.splits if split.account_id == credit_account.id
+        )
+        session.add(activity)
+
+    session.commit()
```
lukasjuhrich, line 77: Please do not commit the session in the `lib`.
While it is true that one should commit often, we cannot judge inside such a function what the lifecycle of the transaction should look like.

For instance, in the following finance endpoint, multiple `lib` functions will be called together:
https://github.com/agdsn/pycroft/blob/865ebcf6bfc5a7a3b02966bf64e00c2342b42717/web/blueprints/finance/__init__.py#L509-L525

You notice that these are lots of operations and only at the end will we judge the transaction to be complete.
Sure, this should be its own lib function, but then the same argument would apply to `end_payment_in_default_memberships`, which cannot know how it will be used.

web/blueprints/finance/__init__.py
```diff
@@ -609,6 +610,8 @@ def bank_account_activities_return_do() -> ResponseReturnValue:
 
         sepa_xml: bytes = generate_activities_return_sepaxml(selected_activities)
 
+        attribute_activities_as_returned(session, selected_activities, current_user)
+
```
lukasjuhrich, line 614: Here is where the commit should happen.

pycroft/model/config.py
```diff
@@ -72,6 +72,11 @@ class Config(IntegerIdModel):
         foreign_keys=[membership_fee_bank_account_id]
     )
 
+    non_attributable_transactions_account_id: Mapped[int] = col(ForeignKey(Account.id))
+    non_attributable_transactions_account: Mapped[Account] = relationship(
+        foreign_keys=[non_attributable_transactions_account_id]
+    )
```
lukasjuhrich, line 78: Remember to add this to the `ConfigFactory`: https://github.com/agdsn/pycroft/blob/865ebcf6bfc5a7a3b02966bf64e00c2342b42717/tests/factories/config.py#L36-L38

lukasjuhrich:
Unfortunately I'll have to postpone this until upgrading to py3.12 because of [this line](https://github.com/mdomke/schwifty/blob/44ae8e7629b98a16296b0827ba64b9326dadf25f/pyproject.toml#L29): This non-exact version requirement somehow makes it down to the packaged wheel, which prevents a deterministic install with `pip wheel` (see build log), which we implicitly enable by saving our `requirements*.txt` with hashes.

lukasjuhrich:
FYI I rebased this branch, so make sure to `pull --rebase`.

lukasjuhrich:
> I'll have to postpone this until upgrading to py3.12

Crosslink #757

PR #749, opened 2024-09-17
Mpsk devices

### LINKED ISSUE #744: Add support for MPSK
In order for MPSK to be used by our members, an appropriate frontend, including an API endpoint, must be implemented.

The following SQL schema is currently planned:
```sql
CREATE TABLE wifi_mpsk_clients(
    owner_id INT NOT NULL,
    name character varying,
    mac macaddr NOT NULL,
    UNIQUE(mac),
    CONSTRAINT fk_owner_id FOREIGN KEY (owner_id) REFERENCES pycroft."user"(id) ON DELETE CASCADE
);
```

A MAC address must be uniquely assignable to a user. However, a user can create as many MAC addresses as they wish.

To better distinguish between devices, it is planned to allow users to assign names to their devices.

The frontend can thus be implemented similarly to the existing interface list. This can then be placed in a separate tab, "WiFi" or similar, on the user page.

To simplify troubleshooting in case of issues, a note should be added in the frontend if the user has a legacy WiFi password that is not stored in plain text. (Reminder: WiFi password ≠ login password)

Feel free to ask questions if necessary.

agmes4:
# Added 
- added MPSK Clients for a user 
- added model representing MPSK Clients 
- added frontend integration for mpsk adding revoming and editing 
- added api endpoints for mpsk 

Resolves #744


pycroft/lib/mpsk_client.py
```diff
@@ -0,0 +1,78 @@
...
+        mac=client.mac,
+    )
+
+    log_user_event(author=processor, user=owner, message=message.to_json())
+
+    return client
+
+
+@with_transaction
+def mpsk_edit(client: MPSKClient, owner: User, name: str, mac: str, processor: User) -> None:
```
lukasjuhrich, line 55: As per [this ADR](https://agdsn.github.io/pycroft/arch/adr-004.html), please pass the `session: Session` as first parameter and then do `session.add(client)` at the end of this function. The same in all the other lib functions you added!
Then, remove all of the `session.add` calls you do in the tests or the API which are superfluous.

_(Ideally, `log_user_event` would already take `session` as its first parameter, but anyway.)_

web/api/v0/__init__.py
```diff
@@ -326,6 +328,85 @@ def get(self, ipv4: IPv4Address | IPv6Address) -> ResponseReturnValue:
...
+        user = get_authenticated_user(user_id, password)
+
+        try:
+            mpsk_client_create(user, mac, name, user)
+            session.session.commit()
+        except InvalidMACAddressException:
+            abort(400, message="Invalid MAC address.")
+        except IntegrityError:
+            abort(400, message="Mac address is already in use.")
+        return "mpsk has been added."
```
lukasjuhrich, line 350: I think it should not be too hard to test the API: we already have frontend tests. So all you need to do is add an `api` package to `tests.frontend` and execute tests in a similar way (all you should need are some fixtures to get a filled database and a `client`).

See https://github.com/flask-restful/flask-restful/pull/110/files for an example of how to use `url_for`, which needs to happen a bit differently because we don't specify endpoint names.

web/api/v0/__init__.py
```diff
@@ -326,6 +328,85 @@ def get(self, ipv4: IPv4Address | IPv6Address) -> ResponseReturnValue:
...
+        },
+        location="form",
+    )
+    def post(
+        self, user_id: int, mpsks_id: int, password: str, mac: str, name: str
+    ) -> ResponseReturnValue:
+        user = get_authenticated_user(user_id, password)
+        mpsk = get_mpsk_client_or_404(mpsks_id)
+
+        if not user == mpsk.owner:
```
lukasjuhrich, line 394:
```suggestion
        if user != mpsk.owner:
```

web/api/v0/__init__.py
lukasjuhrich: You might have to set up an API token to pass it, but that should be the only nontrivial thing.

web/api/v0/__init__.py
```diff
@@ -326,6 +328,85 @@ def get(self, ipv4: IPv4Address | IPv6Address) -> ResponseReturnValue:
...
+    )
+    def post(self, user_id: int, mpsk_id: int, password: str) -> ResponseReturnValue:
+        user = get_authenticated_user(user_id, password)
+        mpsk = get_mpsk_client_or_404(mpsk_id)
+
+        if not user == mpsk.owner:
+            abort(401, message="You are not the owner of the mpsk.")
+
+        mpsk_delete(mpsk)
+        session.session().commit()
```
lukasjuhrich, line 371:
```suggestion
        session.session.commit()
```

web/blueprints/mpskclient/__init__.py
```diff
@@ -0,0 +1,158 @@
...
+        refill_room_data(form, user.room)
+
+    if not form.validate_on_submit():
+        return default_response()
+
+    # existence verified by validator
+    # TODO can't we provide an attribute for that on the form?
+    owner = session.session.get(User, form.owner_id.data)
+    with abort_on_error(default_response), session.session.begin_nested():
+
```
lukasjuhrich, line 63:
```suggestion
```

web/blueprints/mpskclient/__init__.py
```diff
@@ -0,0 +1,158 @@
...
+        return render_template(
+            "generic_form.html", page_title="MPSK Client editieren", form_args=form_args
+        )
+
+    if not form.is_submitted() or not form.validate():
+        return default_response()
+
+    with abort_on_error(default_response), session.session.begin_nested():
+        lib_mpsk.mpsk_edit(mpsk, mpsk.owner, form.name.data, form.mac.data, current_user)
```
lukasjuhrich, line 121: btw, this should be a bug right now, precisely because you don't `session.add(client)` in the `mpsk_edit` definition. See the comment on there.

web/blueprints/mpskclient/__init__.py
```diff
@@ -0,0 +1,158 @@
...
+def user_clients_json(user_id: int) -> ResponseReturnValue:
+    user = get_user_or_404(user_id)
+    # TODO: Importend when the for the returning of actual mpsk mac addresses for MPSK devices
+    # return ""
+    return TableResponse[MPSKRow](
+        items=[_mpsk_row(mpsk, user_id) for mpsk in user.mpsks]
+    ).model_dump()
+
+
+def _mpsk_row(client, user_id: int) -> MPSKRow:
```
lukasjuhrich, line 137:
```suggestion
def _mpsk_row(client: MPSKClient, user_id: int) -> MPSKRow:
```
(and then add an import for `MPSKClient`)

lukasjuhrich (commented):
Overall, these are some high-quality changes! I've found a few details here and there, and some things you could not have known about.

pycroft/model/mpsk_client.py
```diff
@@ -0,0 +1,44 @@
...
+        match = mac_regex.match(mac_address)
+        if not match:
+            raise InvalidMACAddressException("MAC address '" + mac_address + "' is not valid")
+        if int(mac_address[0:2], base=16) & 1:
+            raise MulticastFlagException("Multicast bit set in MAC address")
+        return mac_address
+
+    @validates("name")
+    def validate_name(self, _, name):
+        if len(name.replace(" ", "")) == 0:
```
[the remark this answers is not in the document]
lukasjuhrich: My suggestion was nonsense, I forgot a negation: `if not name.strip(): ...`.
But it is correct that empty strings are `False`y, i.e., it is more idiomatic to check `if [not] <string>` instead of ` if [not] <string> == ""`.

web/blueprints/mpskclient/__init__.py
agmes4: I added an session add so that this should be working

tests/model/test_mpsk.py
```diff
@@ -0,0 +1,81 @@
...
+
+    @pytest.fixture(scope="class")
+    def user(self, class_session):
+        user = factories.UserFactory.build(with_host=True)
+        class_session.add(user)
+        return user
+
+    @pytest.fixture(scope="class")
+    def mpsk_client(self, user):
+        return factories.MPSKFactory()
```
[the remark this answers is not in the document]
agmes4: Thanks will do so :)

ibot3:
Please add a certain limit for the number of MPSK devices per user. (e.g. 50)

agmes4:
I will set it to 10 dont think this should be exceeded by anyone

ibot3:
If you choose a limit this low, there may be situations where it is not enough for a user with many devices.
Maybe add this limitation only on the API side, so that there is no limit for our admins.

lukasjuhrich:
> Maybe add this limitation only on the API side, so that there is no limit for our admins.

Generally in favor, but is there any technical limit we should care about? or do the controllers allow arbitrarily many devices?
This should not matter since MPSK is an edge-case anyway, but I would be curious what the theoretical limit would be; every subnet is finite, anyway.

ibot3:
> Generally in favor, but is there any technical limit we should care about? or do the controllers allow arbitrarily many devices? This should not matter since MPSK is an edge-case anyway, but I would be curious what the theoretical limit would be; every subnet is finite, anyway.

Besides the subnet size, there shouldn't be a limit AFAIK.
I suggested the devices limit mainly to avoid that someone creates lots of devices (maybe hundreds, thousands) with a bad intention.

agmes4:
Anyway I added now the limit of 10 devices we could set it down.

pycroft/lib/mpsk_client.py
```diff
@@ -0,0 +1,81 @@
...
+        log_user_event(message.to_json(), processor, client.owner)
+    return client
+
+
+def mpsk_client_create(owner: User, name: str, mac: str, processor: User) -> MPSKClient:
+
+    if len(owner.mpsks) >= 10:
+        raise AmountExceededError(
+            "the limit of added mpsks clients is exceeded", limit=10, actual=len(owner.mpsks)
+        )
```
ibot3, line 42: Contrary as suggested by me in https://github.com/agdsn/pycroft/pull/749#issuecomment-2360248583, you introduced the limit in the lib.
This also prevents admins to add more devices.

ibot3: This could also be solved by adding an extra parameter to the function that disables the check if needed.

agmes4: Yeah but then I can not test that also I dont think any one would use more then 10 but I can do so

lukasjuhrich: I agree with @ibot3, it is better to have flexibility for the admins. Don't worry about testing it, adding a test for the API can be done later (I will do it).

agmes4: should be In :)

ibot3: I can only speak for me, but I have more than 10 devices at home that only can use PSK wifi.
Especially when using smart-home stuff, this can add-up quite fast.

agmes4: thats true but I dont think that these members are not using a router also maybe its a nice way to get active members :)

agmes4: But I can also set it higher what do you think would be a good number of devices? 50 Sounds a bit much for me in my opinion

ibot3: If the wifi network works as planned in the future, then we don't want that users have their own wifi router, we might even want to prohibit other wifi networks than ours.
However, this is not the case yet.
But our wifi should work in a way that the users don't have the need for another wifi-router.

I would suggest increasing the limit to 30.

agmes4: true I now set it to 30 thank

web/templates/user/user_show.html
```diff
@@ -16,6 +16,12 @@
         'name': 'Hosts',
         'badge': user.hosts | length
     },
+    {
+        'id': 'mpsk',
+        'icon': 'fa-tablet',
+        'name': 'MSPK Clients',
```
ibot3, line 22: Typo

agmes4: Thanks for finding :)

pycroft/lib/mpsk_client.py
lukasjuhrich: > This could also be solved by adding an extra parameter to the function that disables the check if needed.

Sorry for the confusion, but I just now saw that @ibot3 suggested adding a boolean flag. I was agreeing only with his [first message](https://github.com/agdsn/pycroft/pull/749#discussion_r1766468448).

As a rule of thumb, **boolean flags are a [code smell](https://martinfowler.com/bliki/FlagArgument.html)**.

Think about it, this flag makes the problem worse: Instead of removing the check from the lib function, it leaves it in, but _conditionally_. This makes `mpks_client_create` more complex (e.g., it duplicates the test efforts).
This kind of rate limit is not part of `pycroft`s business logic, it's an implementation decision from the API.  Since the API should depend on the lib, but not vice versa, just the fact that the word `api` is now part of a lib function is a clear sign that we're entering the depths of ~~hell~~ accidental complexity.

1. The API endpoint should check it (when we implement it).
2. I also agree with 30, but the fact that we are arguing about the value of a hard-coded number shows that it should be a *configurable environment variable*.

So for now I would ask you to just remove the check, and we can agree on a pre-configured limit of 30 when the API gets implemented.

mabezi:
Looks pretty good to me.

Do we want to add a note in Pycroft if the user has a legacy password? That could simplify debugging, but in the end it affects only very few users.

lukasjuhrich:
@agmes4 I've rebased on top of `develop` and squashed the `!fixup` commits (you can do this yourself with `rebase -i --autosquash`), as you would have done anyway. Just be aware that when pulling, you'll have to rebase on top of the new `origin/MPSK_devices`.

lukasjuhrich:
I'm setting up the API tests. In the meantime, I found a few things to fix.
Most importantly, we're missing an API endpoint to list all of the MPSK devices ;)

web/api/v0/__init__.py
```diff
@@ -343,6 +346,97 @@ def get(self, ipv4: IPv4Address | IPv6Address) -> ResponseReturnValue:
...
+
+        try:
+            # checks rather the user has all settable mpsks clients created
+            if len(user.mpsks) > current_app.config.get("MAX_MPSKS", 30):
+                abort(400, message="User has the maximum count of mpsk clients.")
+
+            if not user.wifi_password:
+                abort(400, message="Legacy wifi password change of password is required.")
+
+            mpsk_client_create(session.session, user, mac, name, user)
```
lukasjuhrich, line 369: @agmes4 BTW, there was a subtle bug here: name and mac are flipped.
For this reason, it is really helpful to always use keyword arguments, even if it seems redundant: `name=name, mac=mac, …`

Im thinking about whether it makes sense to start enforcing [keyword-only arguments](https://docs.python.org/3/tutorial/controlflow.html#keyword-only-arguments).

(I'm on fixing this myself, so click `resolve` if you've read and understood this)

lukasjuhrich:
The API tests have been prepped; @agmes4 continues work and adds an endpoint for the devices.

An important feature has been suggested by @mabezi: when a user has no wifi password or a legacy password, instead of the mpsk table we should add a warning alert "user has no wifi password or a legacy password, and needs a new cleartext password to use MPSK clients".

This is only relevant for edge cases like active members or other folk with `registered_at≤2020`.

web/api/v0/__init__.py
```diff
@@ -354,24 +354,31 @@ class MPSKSClientAddResource(Resource):
...
-            if not user.wifi_password:
-                abort(400, message="Please generate a wifi password first")
+        if not user.wifi_password:
+            abort(412, message="Please generate a wifi password first")
 
-            mpsk_client_create(session.session, owner=user, mac=mac, name=name, processor=user)
+        try:
+            mpsk_client = mpsk_client_create(
+                session.session, owner=user, mac=mac, name=name, processor=user
+            )
```
lukasjuhrich, line 367: Nice! I almost did this myself, but I'm glad you stumbled upon this and noted that the `if`s don't have to be in the `try` block.

lukasjuhrich:
tests (& migration) went through locally, and `ruff` cannot handle a PR of this size, so merging in spite of CI.

PR #753, opened 2024-09-19
Allow memberships beginning in the future to be deleted

### LINKED ISSUE #747: Allow deletion of future group memberships
When planning a group membership that has a start date in the future, there is currently no intuitive way to cancel that membership.
The only workaround existing in the UI is to set both start and end date of the membership in the past. Nonetheless, this will lead to the group membership existing in the group history of the affected user.

rouven0:
Fixes #747

### DIFF
--- a/pycroft/lib/membership.py
+++ b/pycroft/lib/membership.py
@@ -162,6 +162,21 @@ def remove_member_of(
                    user=user, author=processor)
 
 
+def delete_membership(
+    session: Session,
+    membership_id: int,
+    processor: User,
+) -> None:
+    membership = session.get(Membership, membership_id)
+    session.delete(membership)
+    message = deferred_gettext("Deleted membership of  group {group}.")
+    log_user_event(
+        message.format(group=membership.group.name).to_json(),
+        user=membership.user,
+        author=processor,
+    )
+
+
 @with_transaction
 def edit_property_group(
     group: PropertyGroup, name: str, permission_level: int, processor: User
--- a/web/blueprints/user/__init__.py
+++ b/web/blueprints/user/__init__.py
@@ -32,6 +32,7 @@ from flask import (
 )
 from flask.typing import ResponseReturnValue
 from flask_login import current_user
+from flask_wtf import FlaskForm
 from markupsafe import Markup
 
 import pycroft.lib.search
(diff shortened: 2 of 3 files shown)


web/blueprints/user/__init__.py
```diff
@@ -748,6 +768,57 @@ def default_response(refill_form_data: bool = False) -> ResponseReturnValue:
...
+        return default_response()
+
+    with abort_on_error(default_response), session.session.begin_nested():
+        # lib_host.host_delete(host, current_user)
+        delete_membership(
+            session=session,
+            membership_id=membership_id,
+            processor=current_user,
+        )
+        ...
```
lukasjuhrich, line 815: superfluous placeholder?

pycroft/lib/membership.py
```diff
@@ -162,6 +162,22 @@ def remove_member_of(
                    user=user, author=processor)
 
 
+@with_transaction
```
lukasjuhrich, line 165: Remove the `@with_transaction` please, this is an ancient relic – and not necessary, since you correctly took care to `session.commit()` in the `web/` endpoint which consumes this function.

web/blueprints/user/__init__.py
```diff
@@ -748,6 +768,57 @@ def default_response(refill_form_data: bool = False) -> ResponseReturnValue:
...
+        return render_template('generic_form.html',
+                               page_title=("Mitgliedschaft {} für "
+                                           "{} löschen".format(
+                                                membership.group.name,
+                                                membership.user.name)),
+                               membership_id=membership_id,
+                               user=membership.user,
+                               form=form,
+                               form_args=form_args
+                               )
```
lukasjuhrich, line 803: Some formatting concerns I have here can be fixed easily by applying `darker`.
Did you do `pre-commit install`? (the setup script should have done it for you)

web/blueprints/user/__init__.py
```diff
@@ -748,6 +768,57 @@ def default_response(refill_form_data: bool = False) -> ResponseReturnValue:
...
+
+    if not form.is_submitted():
+        return default_response()
+
+    with abort_on_error(default_response), session.session.begin_nested():
+        # lib_host.host_delete(host, current_user)
```
lukasjuhrich, line 809: superfluous comment

rouven0: Removed.

web/blueprints/user/__init__.py
rouven0: Ran darker manually on it. Only did reformatting after I moved the first line.

pycroft/lib/membership.py
rouven0: Removed

web/blueprints/user/__init__.py
rouven0: Removed it.

web/blueprints/user/__init__.py
rouven0: Nevermind, the pre-commit hook seems to be broken on my end and darker still complains in the GitHub checks.

rouven0: Fixed now

PR #758, opened 2024-10-02
Adjust unconfirmed transactions table

FestplattenSchnitzel:
id, author, date and amount are usually not of interest in the first
place.

They're easily accessible still though:
* id: in url of description, e.g. visible via hover
* author: in transaction detail view reachable via click on transaction description
* date: in transaction detail view reachable via click on transaction description
* amount: in bank account activity view reachable via click on bank account activity button

The other bank account owner has been added to the table as that's what's usually checked.

### DIFF
--- a/web/blueprints/finance/__init__.py
+++ b/web/blueprints/finance/__init__.py
@@ -1181,13 +1181,13 @@ def _format_transaction_row(
     bank_acc_act: BankAccountActivity | None,
 ) -> UnconfirmedTransactionsRow:
     return UnconfirmedTransactionsRow(
-        id=transaction.id,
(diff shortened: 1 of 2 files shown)


agmes4:
Looks for sure way slicer but I think the value of the transaction is important to see in directly in order to get a good overview fast also to get immediate attention to important ones.

agmes4 (changes requested):
I think to see the money can be quite handy from time to time

lukasjuhrich:
What's the resolution here?
@FestplattenSchnitzel what does your workflow usually look like – how do you determine what to confirm or not?

PR #762, opened 2024-10-05
Python 3.12

### LINKED ISSUE #757: Upgrade to py3.12
py3.12 has been out for a while now, with a few [highlights](https://github.com/mdomke/schwifty/blob/44ae8e7629b98a16296b0827ba64b9326dadf25f/pyproject.toml#L29):

- [PEP 695](https://docs.python.org/3.12/whatsnew/3.12.html#whatsnew312-pep695), type parameter syntax and the [type](https://docs.python.org/3.12/reference/simple_stmts.html#type) statement
- [PEP 709](https://docs.python.org/3.12/whatsnew/3.12.html#whatsnew312-pep709), comprehension inlining
- [PEP 692](https://docs.python.org/3.12/whatsnew/3.12.html#whatsnew312-pep692), using [TypedDict](https://docs.python.org/3.12/library/typing.html#typing.TypedDict) to annotate [**kwargs](https://docs.python.org/3.12/glossary.html#term-argument)
- [PEP 698](https://docs.python.org/3.12/whatsnew/3.12.html#whatsnew312-pep698), [typing.override()](https://docs.python.org/3.12/library/typing.html#typing.override) decorator

Also, this removes `distutils`, which means we probably need to fix a few of the badly maintained dependencies.

This is a blocker for #739.

lukasjuhrich:
Fixes #757
Required for #739, let's see if the tests pass or semgrep fails.


lukasjuhrich:
Okay, so for mypy I forgot an option in `pyproject.toml` (and we'll have to bump the version), and for semgrep we'll have to disable the rules until https://github.com/semgrep/semgrep/issues/10569 gets fixed.

pycroft/lib/user_deletion.py
```diff
@@ -26,12 +26,8 @@ class ArchivableMemberInfo(Protocol):
     mem_end: datetime
 
 
-TP = t.TypeVar("TP")
-TO = t.TypeVar("TO")
-
-
 # mrh, not available in py3.10…
```
FestplattenSchnitzel, line 29: It's Python 3.12 now so can be removed? ;)

lukasjuhrich: Not sure what this comment refers to.
I would leave it in as a "please check whether everything makes sense" marker for when work on user deletion continues.

lukasjuhrich:
@FestplattenSchnitzel please review and merge if this works on your machine! Or assign somebody else if you don't have time. Given the setup flakiness I'd like a separate pair of eyes.

FestplattenSchnitzel (approved):
Works well on my machine.

PR #769, opened 2025-01-13
Prevent assigning of multiple IP addresses to interface

FestplattenSchnitzel:


ibot3:
Whats the reason for the change?

FestplattenSchnitzel:
When multiple IP addresses in the same subnet are assigned to an interface, `hades-refresh` fails as the table / view constraints are different:

```
Error unknown.IntegrityError: (psycopg2.errors.UniqueViolation) duplicate key value violates unique constraint "auth_dhcp_host_MAC_idx"
DETAIL:  Key ("MAC")=(35:7b:12:48:de:01) already exists.
CONTEXT:  SQL statement "INSERT INTO public.auth_dhcp_host SELECT (diff.newdata).* FROM pg_temp_8.pg_temp_16414_2 diff WHERE tid IS NULL"
[SQL: REFRESH MATERIALIZED VIEW CONCURRENTLY auth_dhcp_host]
(Background on this error at: http://sqlalche.me/e/13/gkpj)
```

PR #770, opened 2025-01-18
fixed sending mpsk clients when loading user data

agmes4:
until now the mpsk clients would not been send when loading a user via the API


hw0lff:
LGTM

PR #773, opened 2025-02-09
Bug fix/mpsk not member

### LINKED ISSUE #772: No MPSK API for not users
# Bug 
currently its possible for users who terminated there account to still setup clients when they can login. That should not be possible 

## Hints 
the function ```has_property()``` can be used for the user to validate if he should be able to setup clients

agmes4:
# Features
Fixes that now users can not add or create mpsk clients
fixed #772

### DIFF
--- a/tests/frontend/api/test_mpsk.py
+++ b/tests/frontend/api/test_mpsk.py
@@ -37,6 +37,11 @@ class TestAddMpsk:
         session.refresh(user)
         assert len(user.mpsk_clients) == 0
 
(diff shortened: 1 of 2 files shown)


ibot3:
Please do not check for the `member` property. There are also users with network access that are not members.

So better check for the `network_access` property.

agmes4:
I see the point but think that this api should just be usable when you are a member. So in case of legacy accounts its still possible to set them when they have wifi access.

agmes4:
never mind looked it up they can login into sipa that was

lukasjuhrich (approved):

PR #777, opened 2025-02-23
Use package groups instead of separate requirements-lockfiles

lukasjuhrich:


lukasjuhrich:
Unfortunately, I forgot that dependabot does not support `uv.lock` yet, see https://github.com/dependabot/dependabot-core/issues/10478. Oh well…

PR #778, opened 2025-02-27
Prevent form from being submitted multiple times by disable button

Crasher508:
By listen on form submission and disable the button, we prevent the ability of sending a request twice

Blocked import button


agmes4:
Looks good to me my be there could also by a visible hint that the system is doing something.

agmes4:
Und my be eine animation könnte die Wartezeit verkürzen

Crasher508:
> Looks good to me my be there could also by a visible hint that the system is doing something.

Your browser show the default page load animation after click the button in the search bar

agmes4:
thats not visible when there are too much taps open

PR #780, opened 2025-06-28
F/added active member api

agmes4:
# Issue 
we need a way to display somewhere if a person is an active member of the AG DSN or not.

# Outcome 
added an API Endpoint for validating for active memberships

### DIFF
--- a/web/api/v0/__init__.py
+++ b/web/api/v0/__init__.py
@@ -60,6 +60,7 @@ from pycroft.model.task import Task
 from pycroft.model.types import IPAddress, InvalidMACAddressException
 from pycroft.model.user import User, IllegalEmailError, IllegalLoginError
 from web.blueprints.mpskclient import get_mpsk_client_or_404
+from web.template_filters import require
 
 api = Api()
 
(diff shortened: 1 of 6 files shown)


web/api/v0/__init__.py
```diff
@@ -940,3 +941,33 @@ def patch(self, token: str, password: str) -> ResponseReturnValue:
...
+        location="form",
+    )
+    def post(self, name: str, byear: int, uid: int) -> ResponseReturnValue:
+        user = session.session.get(User, uid)
+
+        if user is None:
+            return jsonify({"response": False})
+        print(f"bi: {user.birthdate} byear: {byear}")
+        user_status = status(user)
+        has_birthday = False if byear else True
```
ibot3, line 962: `has_birthday = bool(byear)`

web/api/v0/__init__.py
```diff
@@ -940,3 +941,33 @@ def patch(self, token: str, password: str) -> ResponseReturnValue:
...
+class IdentifyResource(Resource):
+    @use_kwargs(
+        {
+            "name": fields.Str(required=True),
+            "byear": fields.Int(required=False, missing=None),
+            "uid": fields.Int(required=True),
+        },
+        location="form",
+    )
+    def post(self, name: str, byear: int, uid: int) -> ResponseReturnValue:
```
ibot3, line 955: I would suggest an `get` endpoint for that, as nothing is created/changed by it.

web/api/v0/__init__.py
```diff
@@ -940,3 +941,33 @@ def patch(self, token: str, password: str) -> ResponseReturnValue:
...
+            if user.birthdate.year == byear:
+                has_birthday = True
+
+        if user_status.is_active and name == user.name and has_birthday:
+            return jsonify({"response": True})
+
+        return jsonify({"response": False})
+
+
+api.add_resource(IdentifyResource, "/user/active")
```
ibot3, line 973: The resource name and path do not really match.
Maybe call it `UserActiveResource`?

agmes4: sounds good for me totally forget to change thanks

web/api/v0/__init__.py
agmes4: nice did not know this thanks just added a not before bool to keep the logic

.idea/runConfigurations/_template__of_py_test.xml
```diff
@@ -1,26 +0,0 @@
...
-      </option>
-    </PathMappingSettings>
-    <option name="_new_keywords" value="&quot;&quot;" />
-    <option name="_new_parameters" value="&quot;&quot;" />
-    <option name="_new_additionalArguments" value="&quot;&quot;" />
-    <option name="_new_target" value="&quot;&quot;" />
-    <option name="_new_targetType" value="&quot;PYTHON&quot;" />
-    <method v="2" />
-  </configuration>
-</component>
```
ibot3, line 26: Why is this deleted?

ibot3 (commented):
Looks fine, but there is still that intellij runconfig template that has been deleted

mabezi (commented):
Looks good to me.

For the MR in general, it might have been nice to deliver more context, what the idea or the goal is.
This might make the review easier for people who are not directly involved in the topic.

But I think this time it's okay :)

Crasher508 (approved):

PR #788, opened 2025-11-10
Add dormitory option to group mail feature

mabezi:
The title says it all. Dormitories should remain optional so it's still possible to send a mail to all members.

An idea for the future would be to migrate dormitories and groups into multiple-choice fields.

### DIFF
--- a/pycroft/lib/user/mail.py
+++ b/pycroft/lib/user/mail.py
@@ -21,6 +21,7 @@ from pycroft.model.user import (
     PropertyGroup,
     Membership,
 )
+from pycroft.model.facilities import Building, Room
 from pycroft.task import send_mails_async
 
 from .user_id import (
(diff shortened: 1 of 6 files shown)


pycroft/lib/user/mail.py
```diff
@@ -125,14 +126,24 @@ def user_send_mail(
...
 
-def group_send_mail(group: PropertyGroup, subject: str, body_plain: str) -> None:
-    users = get_active_users(session=session.session, group=group)
+    return session.scalars(statement)
+
+
+def group_send_mail(
+    group: PropertyGroup, subject: str, body_plain: str, building: Building | None = None
+) -> None:
+    users = get_active_users(session=session.session, group=group, building=building)
```
agmes4:
```suggestion
    users = get_active_users(session=session.session, group=group, buildings=buildings)
```

web/blueprints/user/__init__.py
```diff
@@ -1543,6 +1543,7 @@ def mail_group() -> ResponseReturnValue:
             group=form.group.data,
             subject=form.subject.data,
             body_plain=form.body_plain.data,
+            building=form.building.data if getattr(form, "building", None) else None,
```
agmes4:
```suggestion
            buildings=form.buildings.data if getattr(form, "buildings", None) else [],
```

agmes4 (changes requested):

agmes4 (approved):

PR #789, opened 2025-11-16
Changes from develop

Crasher508:


tests/lib/user/test_mail.py
```diff
@@ -0,0 +1,89 @@
...
+        address=AddressFactory(),
+        membership__group=config.member_group,
+        with_membership=True,
+    )
+
+
+@pytest.fixture(scope="module")
+def building_list(module_session) -> list[Building]:
+    out = []
+    for _ in range(20):
```
lukasjuhrich: I know I'm late to this party, but you can use `BuildingFactory.create_batch(10, …)`

pycroft/lib/user/mail.py
```diff
@@ -125,14 +126,35 @@ def user_send_mail(
     user_send_mails([user], template, soft_fail, use_internal, **kwargs)
 
 
-def get_active_users(session: Session, group: PropertyGroup) -> ScalarResult[User]:
-    return session.scalars(
-        select(User).join(User.current_memberships).where(Membership.group == group).distinct()
-    )
+def get_active_users_with_building(
+    session: Session, groups: t.List[PropertyGroup], buildings: t.List[Building]
```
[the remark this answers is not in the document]
agmes4: I know but I dont like to change it in one file and all the other use the old syntax

web/api/v0/__init__.py
```diff
@@ -940,3 +941,50 @@ def patch(self, token: str, password: str) -> ResponseReturnValue:
...
+        Returns:
+            ResponseReturnValue: A JSON-encoded string with {"response": True} if the user belongs to the
+                     `active_member` group, otherwise {"response": False}.
+        """
+        user = session.session.get(User, uid)
+
+        if user is None:
+            return jsonify({"response": False})
+        user_status = status(user)
+        has_birthday = not bool(byear)
```
lukasjuhrich: Is this a bug?
Why would byear>0 result in has_birthday False, and byear None in True?

tests/lib/user/test_mail.py
agmes4: yeah can change that

web/api/v0/__init__.py
```diff
@@ -940,3 +941,50 @@ def patch(self, token: str, password: str) -> ResponseReturnValue:
...
+    @use_kwargs(
+        {
+            "name": fields.Str(required=True),
+            "fname": fields.Str(required=False, missing=None),
+            "byear": fields.Int(required=False, missing=None),
+            "uid": fields.Int(required=True),
+        },
+        location="form",
+    )
+    def post(self, name: str, fname: str, byear: int, uid: int) -> ResponseReturnValue:
```
lukasjuhrich: The type hints dont match the webargs annotation, the non-required ones should be | None

web/api/v0/__init__.py
agmes4: We agreed on making byear optional

lukasjuhrich:
None of the comments I made above need to be addressed in retrospect, they are purely informative :)

agmes4:
yeah but are good I just merged them cause I thought no one else would check the code as one was very old

PR #797, opened 2026-03-22
Enable python-fints CAMT.053 support

### LINKED ISSUE #796: Replace MT940 with CAMT.053-Support
MT940 is no longer supported. Therefore, it doesn't make sense for us to keep this implementation.

The official replacement is camt.053

FestplattenSchnitzel:
This required dropping our MT940 error handling.
Also drop manual MT940 import capabilities.

Fixes #796


lukasjuhrich (approved):

FestplattenSchnitzel:
@lukasjuhrich do you have an idea how to fix the mypy error at https://github.com/agdsn/pycroft/actions/runs/23392763383/job/68050308316#step:7:14 ?

> pycroft/lib/finance/fints.py:27: error: Returning Any from function declared to return "list[Transaction | Any]"  [no-any-return]

pycroft/lib/finance/fints.py
```diff
@@ -2,50 +2,26 @@
...
+    fints_client: FinTS3PinTanClient,
+) -> list[MT940Transaction | FinTSTransaction]:
     acc = next(
         (a for a in fints_client.get_sepa_accounts() if a.iban == bank_account.iban),
         None,
     )
     if acc is None:
         raise KeyError(f"BankAccount with IBAN {bank_account.iban} not found.")
-    return fints_client.get_filtered_transactions(acc, start_date, end_date)
+    return fints_client.get_transactions(acc, start_date, end_date)
```
lukasjuhrich, line 27: @FestplattenSchnitzel since now we use a (poorly typed) method we do not own, namely `FinTS3PinTanClient.get_transactions`, you have to choose between `cast`ing or using [stubgen](https://mypy.readthedocs.io/en/stable/stubgen.html) and strengthening the info in the `.pyi` stub.

pycroft/external_services/fints.py
lukasjuhrich, line 1: Why did you remove this from `external_services`?

web/blueprints/finance/__init__.py
```diff
@@ -317,46 +309,30 @@ def get_signer() -> Signer:
...
+    assert isinstance(
+        response, NeedTANResponse
+    ), "reconstructed NeedRetryResponse is not an instance of NeedTANResponse"
+
     with client.resume_dialog(dialog_data):
-        client.send_tan(NeedRetryResponse.from_data(challenge), form.tan.data)
+        client.send_tan(response, form.tan.data)
         dialog = client.pause_dialog()
 
     return client, dialog
```
lukasjuhrich: I'm not sure we should keep calling it `dialog` when it's instead the compressed pickled state of the dialog returned by `FinTSDialog.pause()`

lukasjuhrich (approved):

pycroft/external_services/fints.py
lukasjuhrich: cleared orally: because the wrapper client isn't even necessary anymore, so the file became empty.

web/blueprints/finance/__init__.py
FestplattenSchnitzel: Agreed, renamed to `dialog_data`.

PR #798, opened 2026-04-01
OpenID Connect login

Crasher508:
This PR adds the ability to login using OpenID connect
Bildschirmfoto vom 2026-04-01 18-14-34

new dependency: https://pypi.org/project/flask-oidc/


web/blueprints/login/__init__.py
```diff
@@ -54,12 +58,22 @@ def login() -> ResponseValue:
             flash("Erfolgreich angemeldet.", "success")
             return redirect(request.args.get("next") or url_for("user.overview"))
         flash("Benutzername und/oder Passwort falsch", "error")
+    if oidc.user_loggedin:
+        info = flask_session["oidc_auth_profile"]
+        username = info.get("preferred_username")
```
lukasjuhrich: Unfortunately I have no idea what the shape of `oidc_auth_profile` is.
Our current mappings in keycloak are like this:

Image

I am not sure if `preferred_username` is some special thing always provided by keycloak or flask-oidc.

In any case, the intended functionality is:
1. if `pycroft_login` is set, use that. If the corresponding user does  not exist, abort with an error and do not attempt to fall back.
2. if not set, use the `username`.

web/blueprints/login/__init__.py
```diff
@@ -54,12 +58,22 @@ def login() -> ResponseValue:
...
+        user = User.get(username)
+        if info is not None and username is not None and user is not None:
+            login_user(user)
```
lukasjuhrich: Also, you need to verify the condition that the user has the `Mitgliederverwalter` role set: Not all users of the `internal` realm (i.e., all active members) are allowed to log in to pycroft.
this is mapped in keycloak, but I do not know how it is exposed in the user profile.

pycroft/model/user.py
```diff
@@ -395,6 +395,15 @@ def wifi_password(self, value):
...
         return self.wifi_passwd_hash is not None
 
+    @staticmethod
+    def get(login: str) -> User | None:
+        try:
+            user = User.q.filter(User.login == login).one()
+        except NoResultFound:
+            return None
+        else:
+            return user
```
lukasjuhrich: Please do 
```
return session.scalar(
    select(User).where(User.login == login)
)
```
instead, the `.q` and query syntax in general is deprecated

Crasher508: pycroft.model.session.session is deprecated and  sqlalchemy.orm.session.object_session(self) can not be used since get() is a static method.

```
@staticmethod
def get(session: Session, login: str) -> User | None:
       try:
            user = session.scalar(select(User).where(User.login == login))
     ...
```
also need deprecated pycroft.model.session.session in the blueprint. Is there a solution? `verify_and_get()` has the same problem

lukasjuhrich: At some point you might have to use the `session` context local from `session.py`, but the point is that most of the chain you should pass the session explicitly.

So in this case, instead of using `User.get(login)` at the login blueprint, we should do `User.get(login, session.session)`. See a few of the other blueprints who also look up the session from the proxy object but pass it explicitly.

There are a few ways to avoid this “global variable”, but the cleanest approach would be to just stop using flask and migrate to fastapi ;)

See https://agdsn.github.io/pycroft/arch/adr-004.html#decision for a more detailed discussion.

Crasher508:
Last commit add keycloak test-app with config to test/dev container. Setup needs an entry to /etc/hosts:
`127.0.0.1   keycloak`

lukasjuhrich (approved):

PR #801, opened 2026-04-09
Add enpoints to generate bank transfer xml files

Crasher508:
Generate xml-files for:
- user retransfer ("Rücküberweisung")
- invoice payment

Why? Our new bank sometimes loses the payment reference in manual transfers. This error does not occur with an uploaded transfer.

### DIFF
--- a/tests/lib/test_finance.py
+++ b/tests/lib/test_finance.py
@@ -775,6 +775,4 @@ class TestTransfer:
         assert sepa_xml is not None
         assert len(sepa_xml) > 0
         with pytest.raises(ValueError):
-            generate_transfer_sepaxml(
-                bank_account, "owner", "2DE61", "OSDDDE81XXX", "test", 10
-            )
+            generate_transfer_sepaxml(bank_account, "owner", "2DE61", "OSDDDE81XXX", "test", 10)
--- a/tests/lib/test_finance.py
+++ b/tests/lib/test_finance.py
@@ -21,6 +21,7 @@ from pycroft.lib.finance import (
     take_actions_for_payment_in_default_users,
     get_activities_to_return,
     generate_activities_return_sepaxml,
+    generate_transfer_sepaxml,
 )
 from pycroft.model.finance import (
     Transaction,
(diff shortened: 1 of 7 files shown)


tests/lib/test_finance.py
```diff
@@ -762,3 +764,11 @@ def test_generate_sepa_xml(self, session: Session, utcnow):
...
 
         generate_activities_return_sepaxml(get_activities_to_return(session))
+
+
+class TestTransfer:
+    def test_generate_transfer_sepa_xml(self, session: Session, utcnow):
+        bank_account = BankAccountFactory.build(iban="DE61850503003120219540")
+        generate_transfer_sepaxml(
+            bank_account, "owner", "DE61850503003120219540", "OSDDDE81XXX", "test", 10
+        )
```
lukasjuhrich, line 774: You're missing an assertion here

web/blueprints/finance/forms.py
```diff
@@ -174,6 +175,66 @@ def validate_splits(self, field: FormField) -> None:
...
+        except ValueError as err:
+            raise ValidationError(gettext("Invalid IBAN.")) from err
+
+    def validate_bic(self, field: Field) -> None:
+        try:
+            BIC(field.data)
+        except ValueError as err:
+            raise ValidationError(gettext("Invalid BIC.")) from err
+
+    def validate_amount(self, field: Field) -> None:
```
agmes4: The amount should also not be negative

web/blueprints/finance/forms.py
```diff
@@ -174,6 +175,66 @@ def validate_splits(self, field: FormField) -> None:
             raise ValidationError("Buchung ist nicht ausgeglichen.")
 
 
+class BankAccountTransferForm(Form):
```
agmes4, line 178: Do we need two distinct form types? A decorator or composite pattern might be a better solution. We could introduce two classes that act as decorators, referencing a base form and delegating to subclasses for specific reference. This approach would minimize the number of base form classes while still allowing for flexible customization and the creation of variations without constantly adding new form types.

agmes4: Also lowers amount of duplicated code

agmes4 (approved):

PR #803, opened 2026-04-18
Patch retransfer behavior

Crasher508:
- use type2 user id with checksum instead of plain user id for auto assignment
- flip the amount of user retransfer

### DIFF
--- a/web/blueprints/finance/__init__.py
+++ b/web/blueprints/finance/__init__.py
@@ -791,9 +791,10 @@ def bank_account_transfer(user_id: int) -> ResponseReturnValue:
         form = BankAccountIssueTransferForm()
     else:
         user = _get_or_404(session, User, user_id)
-        reference: str = f"{user.id} Rückerstattung zu viel gezahlte Beiträge"
+        reference: str = f"{encode_type2_user_id(user.id)} Rückerstattung zu viel gezahlte Beiträge"
+        # flip the user account balance since we decrease our debts to the user
         form = BankAccountTransferForm(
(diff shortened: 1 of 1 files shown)


web/blueprints/finance/__init__.py
```diff
@@ -791,9 +791,10 @@ def bank_account_transfer(user_id: int) -> ResponseReturnValue:
         form = BankAccountIssueTransferForm()
     else:
         user = _get_or_404(session, User, user_id)
-        reference: str = f"{user.id} Rückerstattung zu viel gezahlte Beiträge"
+        type2_user_id = encode_type2_user_id(user.id)
+        reference: str = f"{type2_user_id} Rückerstattung zu viel gezahlte Beiträge"
```
agmes4: you could also use `reference: str = f"{encode_type2_user_id(user.id)} Rückerstattung zu viel gezahlte Beiträge"`

web/blueprints/finance/__init__.py
```diff
@@ -791,9 +791,10 @@ def bank_account_transfer(user_id: int) -> ResponseReturnValue:
...
         form = BankAccountTransferForm(
-            owner=user.name, reference=reference, amount=user.account.balance
+            owner=user.name, reference=reference, amount=-user.account.balance
```
agmes4: please add a comment why this is necessary for a better readability

agmes4 (changes requested):

agmes4:
Also probably the check for the amount in `BankAccountTransferForm` needs to be changed then

agmes4 (approved):
Looks good

PR #804, opened 2026-07-23
Feature/retransmission

agmes4:
In order to automate retransmission demands from members a workflow implementation for pycroft is needed. 
I implemented an enhanced workflow allowing automatic validation of data and add the api endpoints for sipa


FelixFelix0815 (changes requested):
retransmission->repayment

PR #810, opened 2026-09-16
Feature/verify manual registration

Timaeaeh:
implements intention of #808


web/blueprints/user/__init__.py
```diff
@@ -1286,6 +1287,33 @@ def member_request_edit(pre_member_id: int) -> ResponseReturnValue:
...
+                            " ".join(names[:i]), " ".join(names[i:]) + ";", form.birthdate.data
+                        )
+                        if swdd_person_id == form.person_id.data:
+                            break
+
+                    if swdd_person_id != form.person_id.data:
+                        form.person_id.errors.append(
+                            "Verifizierung mit dem SWDD fehlgeschlagen!",
+                        )
+
```
agmes4, line 1316: this should be moved to a function in odert to have a test. The test could be implemented as a simple unite test making the code verifiable. 

create a function in `pycroft/lib/member_request.py` returning rather an SWDD was found or not and write some Unittests for matching all 4 cases

agmes4 (changes requested):

